PR #192, opened 2021-11-29
Integrate tube and box orbits in parallel

maindlt:
This is the first version of integrating tube and box orbits in parallel, activated by setting the multiprocessing option `orblibs_in_parallel`.

Benchmark GAMA30980 with 6x5x4x3 orbit library, ncpus=ncpus_weights=8, SplitModelIterator, executed via
`time ./test_LegacyGridSearch`:

`orblibs_in_parallel: True`: real	22m58.832s / user	90m46.510s / sys	1m35.988s
`orblibs_in_parallel: False`: real	26m29.226s / user	88m29.011s / sys	1m38.791s

This is a wall time speed-up of about 13%.

Things to be aware of / issues to discuss:

The model iterator calls `Model.get_orblib()` which - if the orblib does not exist already - (i) calculates the initial conditions, (ii) integrates the orbits and executes triaxmass and triaxmassbin, and (iii) calls `OrbitLibrary.read_losvd_histograms()`. This happens in each of the up to `ncpus` parallel processes. This PR further parallelizes (ii) by splitting it into two parallel processes in `OrbitLibrary.get_orbit_library_par()`.

- Hence, (ii) will be most efficient if `ncpus` is 0.5*(number of physical CPU cores), but then (i) and (iii) will not use all cores. As Sabine indicated this will most likely not be an issue on multi-CPU systems (such as VSC4) because the number of new models is typically much smaller than `ncpus` on a cluster node and therefore the orbit integration will use cores that otherwise would not be used anyway.

- If we merge this PR, how will we deal with the `ncpus` vs. actually available cores in case `orblibs_in_parallel` is active? One way is to mention in the documentation that `ncpus` should be about half of the number of available cores and recommend to increase `ncpus_weights` accordingly (which might confuse users, though). Another way is to use only `ncpus`/2 processes in `SplitModelIterator`’s multiprocessing pool for the orbit library to minimize the risk of multiple integration processes on one core...


maindlt:
Hi @sthater, for your example galaxy NGC4528 the speedup is about 28.5% if the number of CPUs > twice the number of models.

Details: Here is a comparison for different `ncpu` settings in the config file (in each case, SLURM jobs were submitted with `--ntasks-per-node=24` `--ntasks-per-core=2`. As the grid search never added more than 8 models in neither case the parallel tube and box orbit integration hit the limit.

orblibs_in_parallel | ncpus | real | user | sys | real speedup | user speedup | sys speedup
-- | -- | -- | -- | -- | -- | -- | --
False | 24 | 913m10.395s | 3110m14.237s | 3m28.138s | 0.00% | 0.00% | 0.00%
True | 24 | 651m25.639s | 3141m29.543s | 3m13.721s | 28.66% | -1.00% | 6.93%
True | 12 | 654m8.794s | 3151m25.199s | 3m35.775s | 28.37% | -1.32% | -3.67%

Running the test again with `--ntasks-per-node=8` `--ntasks-per-core=2` and `ncpus=4` and `ncpus=8` to determine the performance behavior when orbit integrations need to share a core. Will post the results when done ;-)

sthater:
Thank you, Thomas :) Yes, this is a bit closer to what i had expected! How many models did you run? Could you also put in here the chi2 grid? But I think about 30% speedup is already quite nice anyway :)

maindlt:
Hi Sabine,

I ran exactly your script and config file and varied only `ncpus`. `all_models.ecsv` has 16 models: 1 in iteration 0, 8 in iteration 1 (6 new orblibs), and 7 in iteration 2 (no new orblib). So in the table above the limit of `--ntasks-per-node=24` was never reached (2*6=12 orblibs in parallel), resulting in the ~30% speedup.

To see what happens when the number of 2*orblibs exceed the tasks per node, I changed the SLURM setting to `--ntasks-per-node=8` `--ntasks-per-core=2`:

orblibs_in_parallel | ncpus | real | user | sys | real speedup | user speedup | sys speedup
-- | -- | -- | -- | -- | -- | -- | --
TRUE | 8 | 649m31.653s | 3140m46.035s | 3m32.976s | 28.87% | -0.98% | -2.32%
TRUE | 4 | 940m17.295s | 3085m1.322s | 3m9.882s | -2.97% | 0.81% | 8.77%

This shows that for `ncpus=8`, the 6 tube and box orblibs in iteration 1 will be calculated in parallel (total of 12 parallel processes > 8 available), still resulting in a ~30% speedup (not sure if `--tasks-per-core` is in effect when limiting the tasks per node, though...). In case `ncpus=4`, DYNAMITE will calculate the 6 new orblibs in two consecutive batches, as expected resulting in longer runtime.

So I think we can recommend the `orblibs_in_parallel` setting (I am still running the `orblibs_in_parallel=False` setting for the `ncpus=4` scenario to estimate any speedup there, though).

I'll send you the link to the runs and the chi2 grid per Slack...

Cheers,
Thomas

maindlt:
Hi @sthater,
here is the last comparison for the `ncpus=4` case, which also results in a ~30% speedup for parallel orblibs...



orblibs_in_parallel | ncpus | real | user | sys | real speedup | user speedup | sys speedup
-- | -- | -- | -- | -- | -- | -- | --
FALSE | 4 | 1344m56.267s | 3060m41.447s | 3m9.528s |  0.00% | 0.00% | 0.00%
TRUE | 4 | 940m17.295s | 3085m1.322s | 3m9.882s | 30.09% | -0.79% | -0.19%

sthater:
Thomas, Alice and Sabine tested this pull request throughtly on multiple galaxies and with varying orbit libraries and no problems occured. We do not know what happens if only one orbit library fails, though.

Issue #179, opened 2021-10-28
many small things


sthater:
Just a few things that I noticed while working on FCC 47. I will work on them next week. Just to inform everyone:

* add User-defined vmin/vmax for kinematic maps. Vmin,smax/min can be added in the prompt. 
* Put h3 and h4 to be between -0.15 and 0.15.  ✓
* Weird rounding of numbers happens for the parameters, e.g. 2.4--> 2.399999999. This can lead to the value being outside the search range, although it is actually not  (probably difficult to fix)
* Calculate alternative kin_chi2 from kinematic maps ✓
* Add C-m200 relation as alternative to fitting C ✓
* improve circularity plot. Add Lx, to be chosen when running the plotting routine ✓
* Also grid search sometimes seems to have problems when being run only over ml and mbh ✓
* When programm is interrupted, but tube orbit library is finished. Check if it is finished and if yes, only calculate box orbits now. ✓
* I think "triaxreadparameters" function can be removed from plotter as we read all parameters in with config file.. but need to double check (update 2023-05-18 Thomas: has been checked and removed in one of the past PRs) ✓

Here I probably need help

* Plotting still takes quite a long time. Sometimes one hours. Not sure why. ✓
* Big orbit library 41*11*11 sometimes crashes. memory? cluster? ✓

gsantucci90:
A few more small things to maybe add/implement:

- Add 1 Re to the plots
- save the best-fit values/name of the best-fit model folder somewhere so that it can easily be accessed (for plotting, decomposition, etc)
- Implement calculation of the fraction of orbits (within a specified radius?)
- Store best-fit values/ values at 1 Re for the parameters in a file - especially useful when modelling multiple galaxies at the same time
- Add reduced chi2 value in the plots of the maps

- Speed up calculation of mass profiles for the mass plot - currently it takes very long (depending on the number of models selected for the profiles, it can also take 1 hr)

sthater:
an additional "gridsearch" where exact points are calculated 

e.g. one provides an array of qs and ps and if the corresponding models were not calculated yet, DYNA runs those models

maindlt:
On the first item "add User-defined vmin/vmax for kinematic maps. Vmin,smax/min can be added in the prompt.":

- Add cbar_lims='user' to the arguments of kinematic maps plotting.

PR #194, opened 2021-12-05
The default name of the logfile is now the same as the config file, with its extension replaced by .log.

maindlt:
This is addressing an issue in Sabine's issue "many small things" #179 (but does not close it)

(also fixes a bug that causes a crash in the rare event when reattempting a run that crashed before a model directory was written to the all_models table).


maindlt:
Closing this PR without merging because its purpose has been implemented and resolved by PR #242.

PR #201, opened 2021-12-17
Error handling for failed orblibs

prashjet:
Attempt to degrade more gracefully for errors like:
```
RuntimeError: cmd_orb_start exit code 0.
ERROR. Message: Note: The following floating-point exceptions are
signalling: IEEE_INVALID_FLAG IEEE_DIVIDE_BY_ZERO
STOP No triaxial deprojection possible
```
- previously these would raise a `RuntimeError` causing program to stop
- I've added `try/except` around `mod.get_orblib()` to prevent this

I've also added `try/accept` around :
- deleting directories of failed models, which sometimes failed, since
the `ml` directories did not exist (possibly related to change above)
- plotting routines in model iterator, which sometimes failed


maindlt:
Tested with `test_nnls.py` and several variants of the config file (varying `m-bh` in addition to `ml`, `SplitModelIterator` and `ModelInnerIterator`, all 3 ParameterGenerators).

Also provoked errors causing `chi2 == nan`. In case these errors cause an orblib to fail (i.e., `orblib_done == False`), Dynamite will try to calculate the orblib in a subsequent iteration. If such an orblib occurs multiple times within an earlier iteration, more than one process might execute the legacy Fortran orbitstart.f90 for the same orblib, which leads to a Fortran error (but Dynamite will not crash). Dynamite will report this with a warning issued by the model iterator's `create_and_run_model` method. If this 'pseudo crash' deems ok I think this can be merged. :-)

Please test `test_all.sh` (it's still the old version)

PR #216, opened 2022-04-13
Fix for the tutorial notebook parameter_space.ipynb

maindlt:
This PR fixes the tutorial notebook `parameter_space.ipynb` by explaining and implementing the `Parameter.raw_value` and `Parameter.par_value` properties.

Please review the changed text & code and confirm the notebook executes as expected.

### DIFF
--- a/docs/tutorial_notebooks/parameter_space.ipynb
+++ b/docs/tutorial_notebooks/parameter_space.ipynb
@@ -25,23 +25,25 @@
      "name": "stderr",
      "output_type": "stream",
      "text": [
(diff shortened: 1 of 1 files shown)


leahskusa (approved):
**Running the code:**
The notebook runs smoothly and the errors that were printed out before are not there anymore. 
**raw_value vs. par_value:**
The purpose of this notebook is to explain and explore dynamite's parameter space. The difference between `par.par_value` and `par.raw_value` is now easier to understand. It is also much clearer that some of the values are in a logarithmic scale and how to access the same values in linear scale.

PR #231, opened 2022-06-22
Fixes a bug that may occur in all parameter generators in certain cases

### LINKED ISSUE #230: FullGrid parameter generator does not create all parameter combinations
In some instances, `FullGrid` fails to produce a grid that includes all parameter combinations. In the attached example, `q` has a lo-value of 0.05 and a hi-value of 0.99, but while `p` and `u` behave as expected, only `q=0.15`, `q=0.25`, and `q=0.35` are generated by `FullGrid`.

Find the cause of this bug and fix.

kinchi2_plot

maindlt:
Fixes a bug that may occur in all parameter generators in case that for at least one parameter:

**Either**
`minstep=step` and a rounding error occurs when comparing the actual parameter delta to `minstep`

**and/or**
at the edge of the parameter interval (a) the parameter is clipped to `lo` and/or `hi` _and_ (b) the parameter delta=`minstep` _and_ (c) a rounding error occurs when comparing the actual parameter delta to `minstep`.

The fix involves replacing expressions similar to `abs(raw_value-raw_new) >= max(minstep,eps)` by the safer variant `abs(raw_value-raw_new) > max(minstep,eps) - sys.float_info.epsilon`.

Closes #230

Tested with Sabine's galaxy and `FullGrid` and `LegacyGridSearch`, but a small orbit library.

Sabine, it would be great if you tested this with your galaxy under realistic conditions.

### DIFF
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -4,6 +4,7 @@
 Change Log
 ****************
 
+- Bugfix: Fixed a bug that may cause a crash in case a parameter does not have a minstep value
 - Improvement: DYNAMITE will catch and correct the erroneous parameter generator setting minstep>step by setting minstep=step for non-fixed component parameters
 - Bugfix: Fixed a bug that may occur in the parameter generators (ensures that DYNAMITE creates all possible models)
 
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -4,6 +4,8 @@
 Change Log
 ****************
 
+- Bugfix: Fixed a bug that - in rare cases - may occur in the parameter generators
+
 Version: 3.0
 ================
 
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -4,7 +4,8 @@
 Change Log
 ****************
 
-- Bugfix: Fixed a bug that - in rare cases - may occur in the parameter generators
+- Improvement: DYNAMITE will catch and correct the erroneous parameter generator setting minstep>step by setting minstep=step for non-fixed component parameters
+- Bugfix: Fixed a bug that may occur in the parameter generators (ensures that DYNAMITE creates all possible models)
 
 Version: 3.0
 ================
--- a/dynamite/parameter_space.py
+++ b/dynamite/parameter_space.py
@@ -458,9 +458,9 @@ class ParameterGenerator(object):
         else:
             this_iter = np.max(self.current_models.table['which_iter']) + 1
         # check whether we need to do anything in the first place...
+        newmodels = 0
         if not self.status['stop']:
             self.specific_generate_method(**kw_specific_generate_method)
-            newmodels = 0
             # Add new models to current_models.table
             for m in self.model_list:
                 if self._is_newmodel(m, eps=1e-10):
--- a/dynamite/parameter_space.py
+++ b/dynamite/parameter_space.py
@@ -671,7 +671,7 @@ class LegacyGridSearch(ParameterGenerator):
                     # Explicitly set minstep=0 to allow arbitrarily
                     # small steps, not recommended.
                     self.minstep.append(settings['minstep'] \
-                        if 'minstep' in settings else self.step)
+                        if 'minstep' in settings else settings['step'])
                 else:
                     self.step.append(None)
                     self.minstep.append(None)
@@ -745,7 +745,7 @@ class LegacyGridSearch(ParameterGenerator):
                     for s in [-1, 1]:
                         new_raw_value = np.clip(raw_center + s*step, lo, hi)
                         if abs(new_raw_value-par.raw_value) \
-                           > minstep - sys.float_info.epsilon:
+                           >= minstep - sys.float_info.epsilon:
                             self.new_parset[paridx].raw_value = new_raw_value
                             if self._is_newmodel(self.new_parset, eps=1e-10):
                                 self.model_list.append\
@@ -827,7 +827,7 @@ class GridWalk(ParameterGenerator):
                     self.step.append(settings['step'])
                     # use 'minstep' value if present, otherwise use 'step'
                     self.minstep.append(settings['minstep'] \
-                        if 'minstep' in settings else self.step)
+                        if 'minstep' in settings else settings['step'])
                 else:
                     self.step.append(None)
                     self.minstep.append(None)
@@ -937,7 +937,7 @@ class GridWalk(ParameterGenerator):
             raw_values = []
             # start with lo...
             delta = center[paridx] - self.clip(center[paridx] - step, lo, hi)
-            if abs(delta) > minstep - sys.float_info.epsilon:
+            if abs(delta) >= minstep - sys.float_info.epsilon:
                 raw_values.append(self.clip(center[paridx] - step, lo, hi))
             # now mid... tol(erance) is necessary in case minstep < eps
             if len(raw_values) > 0:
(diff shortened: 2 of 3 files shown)


azocchi:
I tested the gridfix version of DYNAMITE against the current one, and these are my results (and some comments at the end). Please note that in the following I am using a modified version of the plots, to make the points more visible and easy to compare (all points are colored here). 

### 1. First config file
I considered the following values for the ```q```, ```p```, and ```u``` parameters in the config file (all the other parameters have ```fixed: True```).
```
q: # intrinsic flattening (C/A)
    par_generator_settings:
        lo: 0.05
        hi: 0.99
        step: 0.1
        minstep: 0.1 
    fixed: False
    value: 0.35 
    LaTeX: "$q_{min}$"
p: # intrinsic B/A
    par_generator_settings:
        lo: 0.5
        hi: 0.999
        step: 0.1
        minstep: 0.01
    fixed: False
    value: 0.79 
    LaTeX: "$p_{min}$"
u: # sigma_obs / sigma_intrinsic
    par_generator_settings:
        lo: 0.80
        hi: 0.9999 
        step: 0.05
        minstep: 0.01
    fixed: False
    value: 0.90 
    LaTeX: "$u_{min}$"
```

When running this with the **current version** of the code, 36 models are calculated. The values sampled for each of the parameters are: 
   *  ```q = [0.15, 0.25, 0.35]```
   *  ```p = [0.69, 0.79, 0.89, 0.99]```
   *  ```u = [0.80, 0.85, 0.90, 0.95, 0.9999]```
plot_orig

When running the same thing with the **gridfix version** of the code, 83 models are calculated. The values sampled for each of the parameters are: 
   *  ```q = [0.05, 0.15, 0.25, 0.35, 0.45, 0.55, 0.65, 0.75]```
   *  ```p = [0.69, 0.79, 0.89, 0.99]```
   *  ```u = [0.80, 0.85, 0.90, 0.95, 0.9999]``` 
plot_fix

### 2. Second config file
```
q: # intrinsic flattening (C/A)
    par_generator_settings:
        lo: 0.05
        hi: 0.99
        step: 0.05
        minstep: 0.1 
    fixed: False
    value: 0.2 
    LaTeX: "$q_{min}$"
p: # intrinsic B/A
    par_generator_settings:
        lo: 0.5
        hi: 0.999
        step: 0.05
        minstep: 0.01
    fixed: False
    value: 0.7 
    LaTeX: "$p_{min}$"
u: # sigma_obs / sigma_intrinsic
    par_generator_settings:
        lo: 0.80
        hi: 0.9999 
        step: 0.05
        minstep: 0.01
    fixed: False
    value: 0.90 
    LaTeX: "$u_{min}$"
```
This was actually a mistake, as I forgot to change ```minstep``` when I changed the ```step``` for the parameter ```q```. When running with the **current version** of the code I got this error: 
```
RuntimeError: cmd_orb_start exit code 0. ERROR. Message: Note: The following floating-point exceptions are signalling: IEEE_INVALID_FLAG IEEE_DIVIDE_BY_ZERO
STOP No triaxial deprojection possible
```

When running with the **gridfix version**, 20 models were calculated and I got no errors:
plot_fix2_wrong

### 3. Second config file but with no mistakes
```
q: # intrinsic flattening (C/A)
    par_generator_settings:
        lo: 0.05
        hi: 0.99
        step: 0.05
        minstep: 0.01 
    fixed: False
    value: 0.2 
    LaTeX: "$q_{min}$"
p: # intrinsic B/A
    par_generator_settings:
        lo: 0.5
        hi: 0.999
        step: 0.05
        minstep: 0.01
    fixed: False
    value: 0.7 
    LaTeX: "$p_{min}$"
u: # sigma_obs / sigma_intrinsic
    par_generator_settings:
        lo: 0.80
        hi: 0.9999 
        step: 0.05
        minstep: 0.01
    fixed: False
    value: 0.90 
    LaTeX: "$u_{min}$"
```

When running this with the **current version** of the code, only 1 model is calculated. 
plot_orig2

When running it with the **gridfix version**, 267 models are calculated. The values sampled for each of the parameters are: 
   *  ```q = [0.05, 0.10, 0.15, 0.20, 0.25, 0.30, 0.35, 0.40, 0.45, 0.50, 0.55, 0.60, 0.65, 0.70, 0.75]```
   *  ```p = [0.65, 0.70, 0.75, 0.80, 0.85, 0.90, 0.95, 0.999]```
   *  ```u = [0.80, 0.85, 0.90, 0.95, 0.9999]```
plot_fix2

### Comments
   *  ```q``` is limited to a maximum value that is lower than the one allowed by the ```hi``` keyword in the config file. I guess that this is due to the fact that some values are not allowed in combination with ```p```, is that correct?
   *  ```p``` is also never going down to the value specified in the ```lo``` keyword, possibly for the same reason?
   *  ```u``` is allowed to assume the value 0.9999 even though it is not at a distance of 0.05 from the previous point in the grid, but I guess that this is due to rounding, right?
   *  We should maybe include a check to ensure that ```minstep``` is indeed lower than ```step``` to avoid mistakes like mine above?

maindlt:
Hi @azocchi,

this is great, Alice - many thanks for testing so thoroughly! :-)

> * `q` is limited to a maximum value that is lower than the one allowed by the `hi` keyword in the config file. I guess that this is due to the fact that some values are not allowed in combination with `p`, is that correct?
> * `p` is also never going down to the value specified in the `lo` keyword, possibly for the same reason?

Yes, if the `q`, `p`, and `u` combination fails the checks in `TriaxialVisibleComponent.validate_parset()` (which essentially just calls `TriaxialVisibleComponent.triax_pqu2tpp()`), the corresponding model will not be created. So I think these two observations are ok.

> * `u` is allowed to assume the value 0.9999 even though it is not at a distance of 0.05 from the previous point in the grid, but I guess that this is due to rounding, right?

This is due to how the parameter generators deal with parameter values clipped to the borders of the permissible interval: if `u` is `0.95`, `u+step` is `u+0.05=1.00`. That is clipped to the interval `[0.80,0.9999]`, resulting in `0.9999`. This value is discarded if it differs from the 'previous' `u` value by less than `minstep` (here, `minstep=0.01`). In this case, `abs(0.95-0.9999)=0.0499>minstep`, therefore `u` is allowed to assume the value `0.9999`. So I think this is ok, too.

> * We should maybe include a check to ensure that `minstep` is indeed lower than `step` to avoid mistakes like mine above?

Absolutely, many thanks for checking with the 'mistake' :-) I will commit a change asap ;-)

Cheers,
Thomas

sthater:
Thank you, Alice! This is awesome!

Regarding point 1 and 2: Exactly! You also see that for low p,  q never reaches the value of the hi keyword: This is because the models not calculated can not be deprojected. A rough rule is q<p<u.

maindlt:
Hi there, the last commit includes `ml` in checking for minstep<=step (which I forgot before) and fixes a bug causing a crash when minstep is not specified at all.
It is tested with `test_nnls.py`, if you have time, please test, too...
I think we are getting very close to merging :-)

azocchi:
Hello! :) 

I tested the last commit by using ```test_nnls.py```, as suggested by @maindlt.

First, I included on purpose a value of ```minstep``` larger than the ```step``` in ```ml```, and got the following warning:
```[WARNING] 15:20:24 - dynamite.physical_system.System - System parameter ml's parameter generator settings have minstep > step, setting minstep=step=4.0.```
After this, the run was carried out and completed without problems, and I got the same results I got when running it with a smaller value of ```minstep```!

Then, I have also tried to remove completely the specification of ```minstep``` and got no error  (and also no warnings about this); again, the run was completed with no problems, and producing the same results. 

I think that we are indeed ready to merge this! ;)

azocchi (approved):

PR #235, opened 2022-07-17
Remove unused orblibs

maindlt:
Implemented a first version of AllModels.remove_unused_orblibs():
Removes orbit libraries for 'bad' models.
Frees disk space by deleting data of models far away from the optimum. Selection of data to delete: either the ``n_models_keep`` best models are kept or models within a ``chi2_rel_thresh`` interval of the so-far best (kin)chi2 value.  Will assume ``chi2_rel_thresh=0.1`` if neither of the parameters is given. Will mark a deleted orbit library with ``orblib_done=False`` and ``weights_done=False`` in the all_models table. If an orblib cannot be deleted because it is used by another model, only the nnls data will be deleted, which is marked by ``weights_done=False`` in the all_models table.

@Sabine, I added you as a reviewer because testing will require DYNAMITE runs with many models to test the new method (status: working draft). Hope that's ok ;-)


sthater:
Can be well tested with FullGrid as there are many models outside of the best-fit chi2 here :)

azocchi:
I did a first test on this by starting a new run on my galaxy (in the same way in which I started my normal run before), and then by using the line suggested by @maindlt:
```c.all_models.remove_unused_orblibs(n_models_keep=5)```

This run produced 2 iterations and 11 models. After the end of the run, I launched the command above and I got a new ```all_models.table```, which I am reproducing here by removing the columns with the parameters and by including a final column that explicitly says if the corresponding orblib has been removed:

```
       chi2             kinchi2       orblib_done weights_done all_done which_iter       directory         removed
------------------ ------------------ ----------- ------------ -------- ---------- ----------------------  -------
277556.30799764424 18974.541784788264        True         True     True          0 orblib_000_000/ml7.00/       no
 254225.1640008418   21266.5694286101       False        False     True          1 orblib_001_000/ml7.00/      yes
  265437.568064754 19423.788293708058        True         True     True          1 orblib_001_001/ml7.00/       no
 264951.6352276332 19251.125326423273        True         True     True          1 orblib_001_002/ml7.00/       no
 247117.2166112598 22066.567627741024       False        False     True          1 orblib_001_003/ml7.00/      yes
263904.14658160275 20409.581263563043       False        False     True          1 orblib_001_004/ml7.00/      yes
252440.10657990136  19565.78827153326       False        False     True          1 orblib_001_005/ml7.00/      yes
 262462.5486033372 18419.891740543768        True         True     True          1 orblib_001_006/ml7.00/       no
 230057.5507614324  20262.54373409823       False        False     True          1 orblib_001_007/ml7.00/      yes
 279574.2700320093  20906.91198317943        True        False     True          1 orblib_000_000/ml6.00/      yes
 276335.7151195228 17832.260072291396        True         True     True          1 orblib_000_000/ml8.00/       no
```
By looking at the values of ```kinchi2``` it seems like only the 5 models with the smallest values of chisquare have been retained, whereas the ones with the largest chisquares have been removed, which is what we want.

BUT:
* the run cannot be restarted after removing the orblibs in this way (I get an error when I try to do that)
* when making plots after removing orblibs there are some errors, but I need to look a bit more carefully into these, to understand exactly where is the problem and how to modify the plotting routines (if possible) to fix it.
* I will also try to run the models for longer before removing orblibs, because it is possible that some of the problems occur because not enough orblibs are retained!

azocchi:
I think that I might have found a small bug... When trying to restart the run to compute some more models before trying to remove orblibs (to test what written above), I got the following error:
```FileNotFoundError: output/models/orblib_000_000/ml10.00nn_orbmat.out not found.```

I think that this is due to the missing slash in the path, because the file ```output/models/orblib_000_000/ml10.00/nn_orbmat.out``` does exist! I did some digging and I think that this could be solved by changing:
* line 88 in ```physical_system.py```, from ```self.parameters[0].update(sformat = '01.2f')``` to ```self.parameters[0].update(sformat = '05.2f')```, to make sure that the number is written by using 5 characters (including 2 decimals), in the form ```XX.XX```, with 0 included in the first place if the number is < 10 (so that we have: 10.00, 09.00 and so on)
* line 460 in ```weight_solvers.py```, from ```self.direc_no_ml = directory_with_ml[:-7]``` to ```self.direc_no_ml = directory_with_ml[:-8]``` to account for the fact that the string ```mlXX.XX/``` is composed of 8 characters.

This seems to be working for me (meaning that now I managed to restart my models with no problem, and all the "ml" directories have indeed the form described above). Should I open an issue for this or should I just change it (if you agree of course) in one of the branches that we will merge soon?

maindlt:
Hi @azocchi,
this is a great catch :-) Given that we will merge this branch soon I would suggest you just change it here.
Meanwhile I am digging through the model iterator to make sure there is no hiccup when it encounters models with a deleted orbit library (`orblib_done=False`, `weights_done-=False`, `all_done=True`)...
Cheers, Thomas :-)

azocchi:
Hey @maindlt, thank you very much for the last commit! :) 
Now I do not get any errors when I remove models and restart the run, and it seems to be working very smoothly! :)

The plots are still giving errors, so I will try to look into those next!

(In the meantime I made the changes suggested above to fix the naming of ml directories)

ejlilley:
I feel like it should be possible to run (e.g.)
`c.all_models.remove_unused_orblibs(n_models_keep=5)`
twice in immediate succession, and for the second run to just make no changes (instead of giving an error)?

maindlt:
> I feel like it should be possible to run (e.g.)
> `c.all_models.remove_unused_orblibs(n_models_keep=5)`
> twice in immediate succession, and for the second run to just make no changes (instead of giving an error)?

Thanks for testing this, @ejlilley! The last commit should fix this bug (the reason was only one of two `rmtree` calls was in a try-except...). If you have time please confirm multiple `c.all_models.remove_unused_orblibs(n_models_keep=5)` calls produce warnings but don't crash.
Thanks and cheers, Thomas

maindlt:
Hi @ejlilley,
following up on the directory errors you got in our last hack session: Did you use the `ModelIterator` with `do_dummy_run=True`? That's the only possible issue resulting from staring at the code ;-) Let's follow up tomorrow...
Cheers,
Thomas

maindlt:
Hello @ejlilley,
many thanks for thoroughly testing - got another bug fixed in the last commit (hopefully the last one in this PR...):
When written as .ecsv, string columns in an astropy table forget their length. Upon reading an existing table, the string columns cannot hold more characters than the longest existing string, hence the problems with the missing slash when adding the leading zero to the `mlxx.xx` subdirectories. Stacking an empty, explicitly typed astropy table and the read-in data fixes this...
I tested with a copy of your data on VSC4, please have another look and confirm that the results are as expected ;-)
Cheers,
Thomas

PR #237, opened 2022-07-26
Plots updates

azocchi:
This fixes some issues in the chisquare corner plot pointed out by Sabine. I have also took this chance to remove the ```sauron``` color maps, which were often causing warnings, and I replaced them with one of the ```cmasher``` color maps suggested by Sabine in the past. If you prefer to have a different color map, please let me know and I can update this! :)

### DIFF
--- a/dynamite/plotter.py
+++ b/dynamite/plotter.py
@@ -10,16 +10,16 @@ from scipy.sparse import csr_matrix
 from scipy.sparse.csgraph import maximum_bipartite_matching
 from copy import deepcopy
 import matplotlib as mpl
-from matplotlib.ticker import MaxNLocator
+from matplotlib.ticker import MaxNLocator, FixedLocator
 from matplotlib.ticker import NullFormatter
 import matplotlib.pyplot as plt
-from plotbin import sauron_colormap as pb_sauron_colormap
+#from plotbin import sauron_colormap as pb_sauron_colormap
 from plotbin import display_pixels
 # from loess.loess_2d import loess_2d
 from dynamite import kinematics
 from dynamite import weight_solvers
 from dynamite import physical_system as physys
-
+import cmasher as cmr
 class ReorderLOSVDError(Exception):
     pass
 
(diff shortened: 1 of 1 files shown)


sthater:
I tested the chisquare plot issue and it works now :) Thank you for the great work!

Before

Screenshot 2022-07-25 at 10 56 27


Now

kinchi2_plot

sthater:
Regarding the colourmap: I would like to have a short discussion next week before we merge this pull request :)

"iceburn" seems to be a map that is really close to the maps that we want to use in the end, so it could be good to use these for the V,h3,h4 maps. However, I would like to see some plots first to see what looks best :) and if we have also good contrast.

azocchi:
I updated the color maps, and this is the resulting plot when running ```dev_tests/test_nnls.py```:
kinematic_map_kinset1

I think that we can merge now! :)

gsantucci90:
A super quick comment: it might be worth adding cmasher to the packages installed with dynamite? Or to the list of the required packages? I got an error when I first run this version of dynamite because I didn't have cmasher :)

Also, I really like the new colours!

maindlt:
Hello @gsantucci90, absolutely agree, I think CMasher should be in the requirements list. I added it and opened a pull request for this...
Have a great day, Thomas :-)

sthater:
Test

gsantucci90:
Hi @azocchi and @maindlt! Sorry, another quick comment: is it possible to cut the colorbars so that we don't have the black bits? If the fits are bad, it's hard to distinguish whether the black bits are the ones from the blue or the red. Like here:
kinematic_map_set1

sthater:
We can adjust the colourbars to be from like 5 to 95% I guess. That is similar to what  we did in the velocity dispersion cut :)

gsantucci90:
Me again, sorry! Is it possible to have the same limits for the surface brightness in the data and the model maps as well? I have slightly different ranges
kinematic_map_set1
.

azocchi:
Hi @gsantucci90! Thank you for pointing out the difference in the limits for the surface brightness maps! I fixed this with my latest commit on the branch `plot_updates`, you should be able to test it from there! Let us know if you spot any other issues! :)

PR #238, opened 2022-07-29
Combine first iterations

### LINKED ISSUE #218: Combine first two iterations
In model iteration, the initial (zeroth) iteration evaluates one model only while the following iteration's parameters do not depend on its result. To take advantage of parallelization: change this so that the first two iterations are combined into a list of >1 models.
This needs to work with all `ParameterGenerator`s and `ModelInnerIterator`s.

maindlt:
Creating the model in iteration 0 (the "first" model) will be done together with creating the models in iteration 1. Thereby the first orbit library and weight calculation will span several models, which better utilizes parallel computing.

Tested with `test_nnls.py` (different `n_max_mods` settings) so far.

Closes #218.

### DIFF
--- a/dev_tests/user_test_config_ml.yaml
+++ b/dev_tests/user_test_config_ml.yaml
@@ -181,7 +181,7 @@ parameter_space_settings:
     stopping_criteria:
         min_delta_chi2_abs : 0.5  # absolute "optimality tolerance"
         #min_delta_chi2_rel : 0.05 # relative "optimality tolerance"
-        n_max_mods : 5 #3
+        n_max_mods : 3
         n_max_iter : 10
 
 legacy_settings:
--- a/dev_tests/user_test_config_ml.yaml
+++ b/dev_tests/user_test_config_ml.yaml
@@ -181,7 +181,7 @@ parameter_space_settings:
     stopping_criteria:
         min_delta_chi2_abs : 0.5  # absolute "optimality tolerance"
         #min_delta_chi2_rel : 0.05 # relative "optimality tolerance"
-        n_max_mods : 3
+        n_max_mods : 5 #3
         n_max_iter : 10
 
 legacy_settings:
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -5,6 +5,7 @@ Change Log
 ****************
 
 
+- Improvement: now the models of the first two iterations are computed together, better utilizing parallel computing
 - Improvement: update publication list
 - Bugfix: fixed wrong version number and copyright year in documentation
 
--- a/dynamite/model_iterator.py
+++ b/dynamite/model_iterator.py
@@ -296,10 +296,10 @@ class ModelInnerIterator(object):
         None.
 
         """
-        iteration = self.all_models.table['which_iter'][-1]
         # new orblib model directories
         nodir = ''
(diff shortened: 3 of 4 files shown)


azocchi:
Hello! I tried testing this, and it seems like indeed, the models from the 0-th and 1-st iterations are calculated all at once. However, I am getting this error:
``` ERROR. Message: At line 194 of file orbitstart.f90 (unit = 31) Fortran runtime error: Cannot open file 'datfil/begin.dat': File exists```
because ```orbitstart.f90``` is trying to create a file that is already there. After doing some digging, I think that the problem is that the first two models are writing in the same directory, even though they are different models (see the table below). Does this make sense or am I doing something wrong somewhere else?

```
m-bh a-bh f-dh q-stars p-stars u-stars ml chi2 kinchi2 orblib_done weights_done all_done which_iter directory
22742594.08282324 1 10000.0 0.35 0.79 0.9  7.0 0.0 0.0 False False False 0 orblib_001_000/ml7.00/
22742594.08282324 1 3162.3  0.35 0.79 0.9  7.0 0.0 0.0 False False False 1 orblib_001_000/ml7.00/
22742594.08282324 1 31622.8 0.35 0.79 0.9  7.0 0.0 0.0 False False False 1 orblib_001_001/ml7.00/
22742594.08282324 1 10000.0 0.31 0.79 0.9  7.0 0.0 0.0 False False False 1 orblib_001_002/ml7.00/
22742594.08282324 1 10000.0 0.39 0.79 0.9  7.0 0.0 0.0 False False False 1 orblib_001_003/ml7.00/
22742594.08282324 1 10000.0 0.35 0.77 0.9  7.0 0.0 0.0 False False False 1 orblib_001_004/ml7.00/
22742594.08282324 1 10000.0 0.35 0.81 0.9  7.0 0.0 0.0 False False False 1 orblib_001_005/ml7.00/
22742594.08282324 1 10000.0 0.35 0.79 0.89 7.0 0.0 0.0 False False False 1 orblib_001_006/ml7.00/
22742594.08282324 1 10000.0 0.35 0.79 0.91 7.0 0.0 0.0 False False False 1 orblib_001_007/ml7.00/
22742594.08282324 1 10000.0 0.35 0.79 0.9  6.0 0.0 0.0 False False False 1 orblib_001_000/ml6.00/
22742594.08282324 1 10000.0 0.35 0.79 0.9  8.0 0.0 0.0 False False False 1 orblib_001_000/ml8.00/
```

maindlt:
Hello @azocchi!

Many thanks for the thorough test, documentation, and pointing to the duplicate directory name issue. You were doing everything right, the problem was that DYNAMITE chose the last iteration for constructing the orblib directory names and this breaks when iteration 0 and 1 both produce new models. The last commit fixes this, I tested with `test_nnls.py` (it now creates `orblib_000_xxx` directories instead of `orblib_001_xxx`) and Sabine's galaxy (correctly creates `orblib_000_xxx` for iteration 0 and `orblib_001_xxx` for iteration 1).

If you have time please run another test and then I think we may be ready to merge?

Cheers,
Thomas

azocchi:
Hi @maindlt, thank you for fixing this! :) 
I tested your new commit with ```test_nnls.py``` and with another run on my galaxy, and in both cases everything went smoothly! :)
I think that we are ready to merge!

PR #239, opened 2022-08-03
reattempt_failures does no longer result in an error if the orblib directory does not exist or multiple to-delete models share the same orblib

maindlt:
When restarting a run with `reattempt_failures: True`, models without an orblib get deleted from disk and the all_models table. Due to a bug, this resulted in an error if multiple models shared the same orblib. This PR fixes this by explicitly removing the unique `orblib_xxx_yyy` directories.

Tested with `test_nnls.py` and several reattempt_failures and provoked error settings. Please test with a realistic galaxy...

### DIFF
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -4,6 +4,7 @@
 Change Log
 ****************
 
+- Bugfix: reattempt_failures does no longer result in an error if multiple to-delete models share the same orblib
 - Improvement: made DYNAMITE compatible with more Linux distributions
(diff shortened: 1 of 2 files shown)


gsantucci90 (approved):
I tested it with the two galaxies where I got the error before, and it works great now. I don't get the error for the libraries anymore, the runs go very smoothly.

maindlt:
Many thanks for testing, @gsantucci90 - I am happy the fix works for you :-) I will merge this PR, please let us know in case you encounter any more problems.
All the best, Thomas

PR #241, opened 2022-08-07
Add another chi2 "type": kinmapchi2

maindlt:
Prepared for adding another chi2 "type": kinmapchi2:
- instead of checking for "chi2" and "kinchi2" in multiple locations throughout DYNAMITE, we now use the method Configuration.validate_chi2() all across DYNAMITE
- the all_models table has another column "kinmapchi2"
- kinmapchi2 still has dummy values (0 for LegacyWeightSolver, nan for NNLS) - NOT FUNCTIONAL YET


maindlt:
`kinmapchi2` should now work with `LegacyWeightSolver`, using the "kinchi2" value from the `LegacyWeightSolver.read_chi2()` method.

Tested with the adapted `dev_tests/test_nnls.py`.

@sthater, please test and advise whether the calculation of kinmapchi2 is ok or whether we need to apply GH_sys_err (and if so, how 😉)

sthater:
Thank you, Thomas! I am already extensively testing this request.

Luckily, we do not need to care about GH_sys_err as these additional errors are added to the errors in the nn_kinem file :)

I still need to test a bit more.


One problem could be, that with these errors DYNA is not backwards compatible anymore. It could make sense to have an extra function in DYNA that checks, whether the kinmapchi2 column is in the allmodels-table and if not adds this column for previous calculations. This of course only works, if no orblibs were deleted previously. 

What do you think?

maindlt:
Hello Sabine, I can't wait to hear what the result of your tests will be :-)

Yes, I agree and think it is a good idea to retrofit 'legacy' all_models tables with the kinmapchi2 column. I will implement a prototype, most probably setting kinmapchi2 to `nan` for models with deleted data (`nn_kinem.out` should be sufficient for kinmapchi2).

maindlt:
Hi Sabine,

DYNAMITE will now retrofit existing `all_models` tables with the new column `kinmapchi2`, calculate its values for existing models whenever possible, and save the new table to disk. This happens right after reading the existing table from disk for the first time when instantiating the configuration object with `reset_existing_output=False` (the default).

Details: Upon reading an all_models file, DYNAMITE checks whether the `kinmapchi2` column exists. If it does not exist, DYNAMITE will add the column to the `all_models` table and try to calculate the `kinmapchi2` values for the existing models. The latter can fail if a weight solver different from `LegacyWeightSolver` is used or model data has been deleted from the disk. In that case, `kinmapchi2` will be `nan`.
Connected small change: placed calculating `kinmapchi2` in a separate method `WeightSolver.chi2_kinmap()`.

Tested with `test_nnls.py` and several settings (weight solver, existing/non-existing `kinmapchi2` column). As this is all brand new, please also test in a realistically large setup ;-)

Cheers,
Thomas

sthater:
Dear Thomas, 

I have tested this pull request on long and short runs for three different galaxies and it works well :) 

One question: I could not quite figure out which chi2 is used by the kinematic maps, that are  created during the run (in model iterator, line 90). I would like to double-check this before merging :) Else, everything is great!

sthater:
Okay, I checked and the minimum chi2 of the corresponding whichchi2 is used. So, all great from my side :)

maindlt:
So glad it works well, Sabine :-)

Hm, perhaps we can talk about fixing the plots in the next hack session. Presently, `Plotter` should at least accept `kinmapchi2` as a valid chi2 (validation via `Configuration.validate_chi2()`)...

azocchi (approved):

sthater:
Everything (that I tested) is working!! :)

Thomas, please merge the plotting PR with this one first, as we did three additional commits there.

Thank you!

PR #242, opened 2022-08-10
Avoid log conflicts when multiple DYNAMITE jobs share one directory

maindlt:
Hello @sthater,

following up on your recent issues with reattempt_failures: as a short-term solution to the problem of distorted log files caused by multiple simultaneous DYNAMITE runs started in the same directory, this PR assigns a unique name to the logs by including a timestamp. This will make debugging based on the detailed logs much easier.

Two little features are in this PR:

- The standard name of the DYNAMITE logfile includes a timestamp and is unique now. Therefore, it is safe to execute multiple DYNAMITE runs from the same directory without distorting the logfile.

- As np.datetime64('now','ms') does not return fractional seconds, timestamps in the all_models table will no longer display fractions of a second.

Please have a look and if possible adopt the changes in `DynamiteLogging` for easier debugging of future runs ;-)

Cheers,
Thomas

### DIFF
--- a/dev_tests/test_nnls.py
+++ b/dev_tests/test_nnls.py
@@ -32,7 +32,9 @@ def run_user_test(make_comp=False):
     else:
         file_dir = None
     fname = 'user_test_config_ml.yaml'
-    c = dyn.config_reader.Configuration(fname, reset_logging=True,
+    c = dyn.config_reader.Configuration(fname,
+                                        reset_logging=True,
+                                        user_logfile='test_nnls',
                                         reset_existing_output=True)
 
     # delete previous output if available
--- a/docs/getting_started/code_overview.rst
+++ b/docs/getting_started/code_overview.rst
@@ -159,7 +159,7 @@ To make plots, you can use the Plotter object:
 
   p = dyn.plotting.Plotter(config=c) # make the plotter object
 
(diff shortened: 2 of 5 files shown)


maindlt:
As discussed in today's hack session: The user can now change the logfile when instantiating the configuration object (added a new optional parameter to the config reader and adapted the logfile parameter of DynamiteLogging).

The parameter `user_logfile` (Configuration) or `logfile` (DynamiteLogging) can take the following values:
A string - name of the logfile (`.log` will be appended).
`None` - will not create a logfile.
`False` - will create a UTC-timestamped logfile dynamiteYYMMDD-HHMMSSuuuuuu.log
Default value: 'dynamite' (creates `dynamite.log`).

Also changed `test_nnls.py` which now uses the `user_logfile` parameter to create `test_nnls.log`.

azocchi:
Thank you, @maindlt! I tested all the options and they all work very well! :)
We should also maybe add a couple of sentences [in the docs](https://dynamics.univie.ac.at/dynamite_docs/getting_started/code_overview.html#logging-output) to introduce this option to the users: I can try and edit the docs in this branch, so that we will not forget to do this! And then, I think that we can merge!

maindlt:
Great, @azocchi, I am glad it all works :-)
Yes, if you have time it will be great if you edit the docs to reflect the new logging functionality (I completely forgot about that). Please let me know in case you need any help!

azocchi:
Done! :) Please, take a look and change whatever you do not like, especially if my explanation is not clear enough!

azocchi (approved):
I think that when merging this we can also remove [the other PR dealing with logging files](https://github.com/dynamics-of-stellar-systems/dynamite/pull/194).

maindlt:
Many thanks @azocchi, your explanation is great! I will go ahead, merge this PR, and close #194 without merging it.

PR #243, opened 2022-08-10
Added cmasher to the list of required packages.

maindlt:
Dear @azocchi,

as it affects the plotter, I'd like to ask for your review (not hyper-urgent):

Per Giulia's comment: added `CMasher` to `requirements.txt`. I think it is a good idea to include it in the setup of DYNAMITE.

While not sure which version is required, I added `cmasher>=1.6.0` (1.6.0 is the latest version and has been released a while ago: Apr 6, 2021).

Installation tested on macOS 11.6.8, Python 3.9.13, including varying the required CMasher version (e.g., `cmasher==1.5.0`).

Please test if `setup.py` works, then I think we can merge ;-)

Thanks and cheers,
Thomas

### DIFF
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -4,6 +4,7 @@
 Change Log
 ****************
 
+- Bugfix: included cmasher in the list of required packages
 - Bugfix: reattempt_failures will no longer result in an error if multiple to-delete models share the same orblib or the orblib directory does not exist
 - Improvement: made DYNAMITE compatible with more Linux distributions
 - Improvement: update publication list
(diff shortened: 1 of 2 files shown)


sthater:
I tested this on the cluster and the installation with setup works!  I think this can be merged with version 1.6.0.

Thank you, Thomas!

maindlt:
Thanks, Sabine :-) Given that it is such a small change I will go ahead and merge this PR.

gsantucci90:
Me again, sorry! Is it possible to have the same limits for the surface brightness in the data and the model maps as well? I have slightly different ranges.
kinematic_map_set1

PR #245, opened 2022-08-29
some updates for plotting

azocchi:
The triangle plot is now updated, to fix some issues pointed out by @sthater (Sabine, please take a look and see if all is ok now!). Moreover, the plots are now accepting ```kinmapchi2``` as a valid value for ```which_chi```: this should solve the plotting issues pointed out [here](https://github.com/dynamics-of-stellar-systems/dynamite/pull/241).


sthater:
I added to one-liner changes :)  Tested on my ATLAS3D galaxies. Further testing should not be necessary. Thanks for taking care of this, Alice!

Below, I show the effect of the second commit: 
kinematic_map_set1_test
kinematic_map_set1

PR #248, opened 2022-09-09
Update docs

maindlt:
- Add most current link to Thater et al. 2022 to index page (arXiv link). **DONE**
- Add this paper to "how to cite". **DONE**, Note: the BiBTeX still points to the arXiv entry, need to change to A&A as soon as per is published.
- Add "how to cite" to menu on the left. **DONE**
- Create a scientific introduction to DYNAMITE ("What is DYNAMITE?", plots, etc.).
- DYNA does not work with Python 3.7 and lower anymore! Needs to be changed! **DONE**
- In the documentation, redesign these sections... Especially: Split Installation Guide into 1. System&Software requirements and 2. Installation Guide. Requirements should automatically be added to the documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,18 +6,12 @@ A package for Schwarzschild- and stellar-population modelling of stellar systems
 
 ## Installation
 
-See [the documentation](https://dynamics.univie.ac.at/dynamite_docs/getting_started/installation.html) for full instructions.
-
-There are two steps.
-
-1. Compile and install the Galahad library and the DYNAMITE Fortran executables.
-   - If you are using a modern Linux distribution with gfortran, then follow the instructions in ``legacy_fortran/README.linux``.
-   - If you are using Mac OS X (or Linux with a nonstandard compiler) then
-     1. Install the Galahad optimization library
-       - ``./install_galahad`` in the directory ``legacy_fortran/galahad-2.3/``
-     2. Compile the Fortran programs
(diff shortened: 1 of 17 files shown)


ejlilley:
oops I didn't mean to add those spurious commits from a different branch. Maybe someone with more git knowledge can revert them?

azocchi:
Hello! I have included the updates we discussed for the documentation. 
@sthater: can you please check that everything is ok and that I have not forgotten anything? 
@ejlilley: when Sabine gives us green light, can you please update the documentation on the website? Thanks in advance to you both! 
@maindlt: I would not merge this PR yet, because some of the points in your message above are still not done... what do you think, is it ok if we use this documentation to publish on the website, even though it is different from the one on the master branch? Maybe if we keep track of this it's going to be fine, no?

maindlt:
Hi @azocchi, hm, I have a slight preference for keeping the official doc in sync with master...
Given that it is 'only' documentation which won't break any code I think it would be ok to merge the current state of this (`doc_update`) branch into `master` and leave the PR open for future changes, what do you think?

azocchi:
Ah, sure! I did not realise that this was an option! :) Please go ahead, then!

maindlt:
Waaahhh - I just noticed that an 'intermediate merge' does not exist and closes the PR. I will open another one for the next round of documentation updates...

Issue #256, opened 2022-10-03
Prevent DYNAMITE from repeatedly trying to solve unsolvable models (Meng's issue)


maindlt:
Implement a fix along the lines of
<img width="288" alt="image" src="https://user-images.githubusercontent.com/8792350/193541942-b7e4d961-e573-4514-b5b0-53ff0039534a.png">

momomaycry:
I also find a bug after removing empty orblibs. I add to lines to in the 

```
for row in rows_orblib:
    self.logger.info(f'new orblib {orblib_dir}’)
for row in rows_ml:
   self.logger.info(f'orblib for ml {orblib_dir}')
```

Then I got the following output: 

> [INFO] 10:48:05 - dynamite.model.AllModels - Model directory orblib_022_005/ removed.
> [INFO] 10:48:06 - dynamite.model.AllModels - Model directory orblib_012_002/ removed.
> [INFO] 10:48:06 - dynamite.config_reader.Configuration - Instantiated AllModels object
> [INFO] 10:48:06 - dynamite.model_iterator.ModelIterator - LegacyGridSearch: iteration 35
> [INFO] 10:48:13 - dynamite.parameter_space.LegacyGridSearch - LegacyGridSearch added 36 new model(s) out of 53
> [INFO] 10:48:14 - dynamite.model_iterator.ModelInnerIterator - orblib for ml orblib_019_000
> [INFO] 10:48:14 - dynamite.model_iterator.ModelInnerIterator - orblib for ml orblib_019_000
> [INFO] 10:48:14 - dynamite.model_iterator.ModelInnerIterator - orblib for ml orblib_022_005
> ......
> [INFO] 10:48:14 - dynamite.model_iterator.ModelInnerIterator - orblib for ml orblib_022_005
> [INFO] 10:48:14 - dynamite.model_iterator.ModelInnerIterator - orblib for ml orblib_022_005

orblib_022_005/ was removed, but it didn’t appear as a new orblib, but only as a new ml, therefore several new ml will solve the orblib for orblib_022_005/ together, and caused conflict errors, as I submitted in **Issue 257**
**I suppose it is because the all_models used in self.is_new_orblib() were not updated after removing those empty models.**

> [INFO] 10:52:55 - dynamite.orblib.LegacyOrbitLibrary - Calculating initial conditions
[...]

Issue #257, opened 2022-10-03
Fortran runtime error: End of file


momomaycry:
"""
Traceback (most recent call last):
  File "/home/myang/.linuxbrew/Cellar/python@3.9/3.9.2_4/lib/python3.9/site-packages/multiprocess/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/home/myang/.linuxbrew/Cellar/python@3.9/3.9.2_4/lib/python3.9/site-packages/multiprocess/pool.py", line 48, in mapstar
    return list(map(*args))
  File "/home/myang/dynamite-master/dynamite/model_iterator.py", line 388, in create_and_run_model
    orblib = mod.get_orblib()
  File "/home/myang/dynamite-master/dynamite/model.py", line 668, in get_orblib
    orblib.get_orblib()
  File "/home/myang/dynamite-master/dynamite/orblib.py", line 100, in get_orblib
    self.get_orbit_ics()
  File "/home/myang/dynamite-master/dynamite/orblib.py", line 328, in get_orbit_ics
    raise RuntimeError(text)
RuntimeError: cmd_orb_start exit code 2. ERROR. Message: At line 76 of file iniparam_f.f90 (unit = 13, file = 'infil/parameters_pot.in')
Fortran runtime error: End of file
Logfile: /home/myang/8138-6102/models/orblib_019_000/datfil/orbstart.log.
"""

momomaycry:
It seems to be the consequence of another bug, as I commented in #256

PR #262, opened 2022-10-13
Bugfix for reattempt_failures=False

maindlt:
If `reattempt_failures=False`, in certain cases it could occur that orblibs of successful models were deleted.

### DIFF
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -4,6 +4,7 @@
 Change Log
 ****************
 
+- Bugfix: If reattempt_failures=False, in certain cases it could occur that orblibs of successful models were deleted
 - Bugfix: Fixed a bug related to a nonexistent model directory if a crash occurs between the parameter generator adding a model and starting to solve it
 - Improvement: Dynamite will no longer crash upon Legacy Fortran errors, but issue warnings and assign nan to the affected chi2 values
 - Improvement: When executing a dummy run (do_dummy_run==True), model_iterator will set both kinchi2 and kinmapchi2 to nan (instead of zero)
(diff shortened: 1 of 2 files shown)


maindlt:
Hi @momomaycry,
please test this, it should significantly improve DYNAMITE's behavior with `reattempt_failures: False`...
Many thanks,
Thomas

May close #256 and #257

momomaycry:
Hi @maindlt,

I haven’t test this yet, but I used the try..except in #201 and remove the part of removing failed models (and staging.ecsv), it works now. I believe this change will also fix the bug. 

By the way, I am not sure with try...except in #201 and also the np.nan for failed models, the staging file system is still essential. If the model is fully done, there is a routine to read from the orbit_weights.ecsv.
In case the jobs were killed accidentally, I included the the following function in the ModelInnerIterator to fix a particular iteration, which provides similar function of reattempt_failed_models but don’t grid a new iteration. 
```
def fix_previous_iteration(self, iter):
        rows_to_do = np.where((self.all_models.table['all_done'] == False) & 
                              (self.all_models.table['which_iter'] == iter))
        rows_to_do = rows_to_do[0]
        self.n_to_do = len(rows_to_do)
        rows_to_do_orblib=[i for i in rows_to_do if self.is_new_orblib(i)]
        n_orblib = len(rows_to_do_orblib)
        rows_to_do_ml=[i for i in rows_to_do if i not in rows_to_do_orblib]
        input_list_orblib = list(enumerate(rows_to_do_orblib))
        input_list_ml = list(enumerate(rows_to_do_ml, start=n_orblib))
        with Pool(self.ncpus) as p:
            output_orblib = \
                p.map(self.create_and_run_model, input_list_orblib)
            output_ml = p.map(self.create_and_run_model, input_list_ml)
        # save the output
        self.write_output_to_all_models_table(rows_to_do_orblib,
                                              output_orblib)
        self.write_output_to_all_models_table(rows_to_do_ml, output_ml)
        self.all_models.save() # save all_models table once models are done
        return 1
```

Issue #255, opened 2022-10-02
Add orbit_exploration.py


maindlt:
Add a new DYNAMITE module `orbit_exploration.py` for orbit analysis tools.
Starting point should be @gsantucci90's decomposition analysis.

PR #267, opened 2022-11-14
Add decomposition

### LINKED ISSUE #266: Add decomposition
Add decomposition based on Giulia's ansatz.

maindlt:
Working towards fully integrating @gsantucci90's decomposition into DYNAMITE.

Current status:
- New class `decomposition.Decomp` with methods `comps_aphist` and `plot_comps_giu` (methods and their names not final).
- `Decomp` needs configuration object.
- `Decomp` uses DYNAMITE's folder structure with the decomposition output written to `config.settings.io_settings['output_directory'] + 'figure_nn/'`.
- Example call in `dev_test/test_decomp.py`.

Here's a summary of key additions/changes to "core" DYNAMITE:
- `setup.py` now uses `gausherm.f` for building the Python module `pyfort_GaussHerm` using `f2py`. For this to work, (1) `setuptools` needed to be replaced by `numpy.distutils` and (2) `numpy.distutils`' `install_lib` method needed to be changed in order to preserve the executable flag on the compiled legacy_fortran components.
- New method `AllModels.get_best_n_models_idx`: gets the indices of the `n` best models in `all_models`.

Problem that needs to be fixed which only occurs for `conversion='gh_fit'`:
- Different outcome when using `orblib.LegacyOrbitLibrary.read_losvd_histograms` as opposed to `decomposition.Decomp.read_losvd_histograms`
- Decomp method -> all conversions produce figures that make sense
- DYNAMITE method -> conversion 'gh_fit' figure does not make sense
All other values for `conversion`  seem to work with either `read_losvd_histograms` method.
Introduced a temporary `Decomp` parameter for diagnostics: `read_orblib` which can take values `'decomp'` or `'dynamite'` (use: see `dev_test/test_decomp.py`).

When merged, this PR closes #266.


prashjet:
Confirmed with @gsantucci90 and @adriano-poci that `gh_fit` and `fortran` should be doing similar enough things that we can rely on one for the workshop. So if we can't solve the `gh_fit` problem soon, we should merge this PR already, and then keep the issue alive to solve in the future.

One general change I would suggest before merging is to use more descriptive names for all methods and keyword options, at least the ones that will be user facing. Some suggestions:

- 'losvd_vsig' --> `gh_expand_around_losvd_mean_and_std_deviation`
- 'gh_fit' --> `gh_fit_with_free_v_sigma_params`
- 'fortran'  --> `gh_fit_with_free_v_sigma_params_fortran`
- 'moments' --> ??? :P

gsantucci90:
Should this also close https://github.com/dynamics-of-stellar-systems/dynamite/issues/255 ? I think at the beginning the idea was to call the module orbit_exploration :D

maindlt:
> Should this also close #255 ? I think at the beginning the idea was to call the module orbit_exploration :D

Thanks for reminding us, @gsantucci90! If I remember correctly, we intended  to contain a collection of (future) orbit analysis tools with decomposition being the first one.

If so (and please correct me if I'm wrong!):
- I would propose to rename `decomposition.py` to `orbit_exploration.py` (containing only the `Decomp` class for now which we might even rename to `Decomposition`).
- #255 would be linked to this PR but remain open for future orbit analysis tools?

@gsantucci90: what do you think about naming the module `orbit_exploration` vs. `decomposition`?
People in EU timezone: shall we finalize the naming in our next hack session this Friday?

gsantucci90:
@maindlt I like the name orbit_exploration :)

azocchi:
I have tested this on a couple of galaxies, and it works nicely for me! :)

The only thing that does not work for me, is that I cannot install DYNAMITE on the cluster when I am on this branch, because I am getting the following error:

```
/home/fs71474/zocchia/dyn3/dynamite/setup.py:3: DeprecationWarning: 

  `numpy.distutils` is deprecated since NumPy 1.23.0, as a result
  of the deprecation of `distutils` itself. It will be removed for
  Python >= 3.12. For older Python versions it will remain present.
  It is recommended to use `setuptools < 60.0` for those Python versions.
  For more details, see:
    https://numpy.org/devdocs/reference/distutils_status_migration.html 


  import numpy.distutils.core
Traceback (most recent call last):
  File "/home/fs71474/zocchia/dyn3/dynamite/setup.py", line 3, in <module>
    import numpy.distutils.core
  File "/home/fs71474/zocchia/.local/lib/python3.9/site-packages/numpy/distutils/core.py", line 24, in <module>
    from numpy.distutils.command import config, config_compiler, \
  File "/home/fs71474/zocchia/.local/lib/python3.9/site-packages/numpy/distutils/command/config.py", line 19, in <module>
    from numpy.distutils.mingw32ccompiler import generate_manifest
  File "/home/fs71474/zocchia/.local/lib/python3.9/site-packages/numpy/distutils/mingw32ccompiler.py", line 28, in <module>
    from distutils.msvccompiler import get_build_version as get_build_msvc_version
ModuleNotFoundError: No module named 'distutils.msvccompiler'
```

Everything else works fine, and I have no problems on my laptop, so I think that when we fix this we can safely merge this branch!

maindlt:
Testing update:
- on my laptop (Python 3.9.13) and
- on VSC4 with no modules loaded, but miniconda installed (Python 3.10.8),

installing DYNAMITE and running test_nnls.py and test_decomp.py work for me as expected.
Let's hunt down your error this week ;-)

maindlt:
Will merge with master for expedited use. Will also open an issue for adding `orbit_exploration.py` docstrings and documentation...

PR #277, opened 2023-02-02
Ensure staging files get deleted after successful iteration

maindlt:
Ensure all model inner iterators delete staging files after successfully completing an iteration, which was not the case for the `SplitModelIterator`.

So far successfully tested with `test_nnls.py` and both model inner iterators: staging files are deleted and according info messages are written to the log.

### DIFF
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -4,6 +4,7 @@
 Change Log
 ****************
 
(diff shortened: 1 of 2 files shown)


azocchi:
I did some tests, and all seems perfect to me! Go ahead and merge when you are ready! Thanks!

PR #280, opened 2023-02-19
Orbit plot now works with scipy nnls

maindlt:
`Plotter.orbit_plot()` now reads the orbit weights using standard DYNAMITE methods and does no longer rely on legacy weight solver's output files.

The code has some testing features in it and needs cleanup before merging.

Please test. Hints & expected behavior:

- Execute the model iterator (ested with dev_tests/test_nnls.py so far).
- Create an orbit plot. E.g., via
```
p = dyn.plotter.Plotter(c)
p.orbit_plot(Rmax_arcs=3000)
```
- In case of the legacy weight solver, there should be two identical figures orbit_library_only_LegacyWeightSolverOLD.png and orbit_linear_only_LegacyWeightSolver.png in the plots directory.
In case of the scipy nnls weight solver, there should be one figure orbit_linear_only_NNLS.png in the plots directory.

Many thanks and cheers,
Thomas

### DIFF
--- a/dev_tests/test_nnls.py
+++ b/dev_tests/test_nnls.py
@@ -109,9 +109,12 @@ def run_user_test(make_comp=False):
         print('The best 2 models:')
         c.all_models.get_best_n_models(n=2).pprint(max_lines=-1, max_width=-1)
 
-    return c.all_models.table, \
-        compare_file, \
-        c.settings.parameter_space_settings['stopping_criteria']['n_max_mods']
+    if make_comp:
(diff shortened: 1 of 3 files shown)


azocchi:
Hello! I tested this with two cases and I confirm that I get two identical figures ``orbit_library_only_LegacyWeightSolverOLD.png`` and ``orbit_linear_only_LegacyWeightSolver.png`` in the plots directory. 

I think that the new implementation is all good and you can proceed and merge! :)

PR #281, opened 2023-02-28
Decomposition generalized and made ready for more complex projections

maindlt:
This draft PR serves as a collector for changes to the `Decomposition` class to make it ready for alternative orbit projection.


maindlt:
There seems to be a problem with histogram normalization in some bins in `dyn.kinematics.Histogram.get_mean_sigma_gaussfit()`. As this needs to be attended to I set this PR back to draft...

maindlt:
The problem was actually in fitting a Gaussian to some strangely-shaped losvd histograms. The fix in commit 6a16422 increases the stability of `Histogram.get_mean_sigma_gaussfit()` by

- using a better estimate of the Gaussian amplitude
- catching `nan` values in data
- replacing the Levenberg-Marquardt curve fitting algorithm by the more robust Trust Region Reflective algorithm (see scipy.optimize.least_squares documentation)
- catching (unexpected) `scipy.optimize.curve_fit` errors

Also, `Decomposition` now normalizes the losvd histogram data before determining mean and sigma values (last commit).

So far tested successfully with NGC6278 and FCC047 (2 kinematics sets) data.

azocchi:
Closes #292.

PR #285, opened 2023-03-14
New orb plot

prashjet:
Projection tensor implemented, plotting still outstanding.  In the meantime, please test the current status.

Install [sparse](https://sparse.pydata.org/en/0.14.0/), then:

```
import dynamite as dyn
c = dyn.config_reader.Configuration( ... ) # your config file here
mod = c.all_models.get_model_from_row( ... ) # get any model here
orblib = mod.get_orblib()
orblib.read_orbit_property_file()
```

At this stage, look at `orblib.orb_properties` which should be a table with properties (e.g. distance, angular momenta) of the orbits. Then:

```
orblib.classify_orbits(make_diagnostic_plots=True)
fig = plt.gcf() # save this figure if made successfullly
orblib.get_projection_tensor()
print(orblib.projection_tensor.size)
```

Please add the output, plot, and all lines in the logfile between `Orbit library classification` and `Orbit library classification DONE` here in this thread. Thank you!

### DIFF
--- a/dynamite/orblib.py
+++ b/dynamite/orblib.py
@@ -776,7 +776,7 @@ class LegacyOrbitLibrary(OrbitLibrary):
         orbclass1 = orbclass1.reshape((5, ncol, nrow), order='F')
         orbclass2 = np.genfromtxt(self.mod_dir+'datfil/orblibbox.dat_orbclass.out')
         orbclass2 = orbclass2.ravel()
-        orbclass2 = orbclass1.reshape((5, ncol, nrow), order='F')
+        orbclass2 = orbclass2.reshape((5, ncol, nrow), order='F')
         orbclass=np.dstack((orbclass1,orbclass1,orbclass2))
         orbclass1a=np.copy(orbclass1)
         orbclass1a[0:3,:,:] *= -1
@@ -877,6 +877,7 @@ class LegacyOrbitLibrary(OrbitLibrary):
         self.logger.info(f'    - {percent(n_xt_exact/n_xtish)}% of x-tubes')    
         self.logger.info(f'    - {percent(n_yt_exact/n_ytish)}% of y-tubes')    
         self.logger.info(f'    - {percent(n_zt_exact/n_ztish)}% of z-tubes')
+        self.logger.info('Orbit library classification DONE.')
         # save the output
         orb_classification = {
             'bool_box':bool_box,
--- a/dynamite/orblib.py
+++ b/dynamite/orblib.py
@@ -757,6 +757,28 @@ class LegacyOrbitLibrary(OrbitLibrary):
         ml_original = float((lines[-9])[0])
         return ml_original
 
+    def read_orbit_property_file_base(self, file, nrow, ncol):
+        """
+        read in `datfil/orblib.dat_orbclass.out`
+        
+        which stores the information of all the orbits stored in the orbit library
+        norb = nE * nI2 * nI3 * ndithing^3 
+        for each orbit, the time averaged values are stored:
+        lx, ly ,lz, r = sum(sqrt( average(r^2) )), Vrms^2 = average(vx^2 + vy^2 + vz^2 + 2vx*vy + 2vxvz + 2vxvy)
+        
+        The file was stored by the fortran code orblib_f.f90 integrator_find_orbtype
+        """
+        data=[]
+        lines = [line.rstrip('\n').split() for line in open(file)]
+        i = 0
+        while i < len(lines):
+            for x in lines[i]:
+                data.append(np.double(x))
+            i += 1
+        data=np.array(data)
+        data=data.reshape((long(5),int(ncol),int(nrow)), order='F')
+        return data
+
     def read_orbit_property_file(self):
         """Read the file `*orbclass.out` files
 
@@ -771,12 +793,14 @@ class LegacyOrbitLibrary(OrbitLibrary):
         norb = int(nE*nI2*nI3)
         ncol = ndither**3
         nrow = norb
-        orbclass1 = np.genfromtxt(self.mod_dir+'datfil/orblib.dat_orbclass.out')
-        orbclass1 = orbclass1.ravel()
-        orbclass1 = orbclass1.reshape((5, ncol, nrow), order='F')
-        orbclass2 = np.genfromtxt(self.mod_dir+'datfil/orblibbox.dat_orbclass.out')
-        orbclass2 = orbclass2.ravel()
-        orbclass2 = orbclass2.reshape((5, ncol, nrow), order='F')
+        orbclass1 = self.read_orbit_property_file_base(
+            self.mod_dir+'datfil/orblib.dat_orbclass.out',
+            ncol,
+            nrow)
+        orbclass2 = self.read_orbit_property_file_base(
+            self.mod_dir+'datfil/orblibbox.dat_orbclass.out',
+            ncol,
+            nrow)
         orbclass=np.dstack((orbclass1,orbclass1,orbclass2))
         orbclass1a=np.copy(orbclass1)
         orbclass1a[0:3,:,:] *= -1
--- a/dynamite/orblib.py
+++ b/dynamite/orblib.py
@@ -4,6 +4,8 @@ import shutil
 import logging
 import numpy as np
 from scipy.io import FortranFile
+from astropy import table
+import astropy.units as u
 
 from dynamite import physical_system as physys
 from dynamite import kinematics as dyn_kin
@@ -752,5 +754,44 @@ class LegacyOrbitLibrary(OrbitLibrary):
         lines = [line.rstrip('\n').split() for line in open(infile)]
         ml_original = float((lines[-9])[0])
         return ml_original
+
+    def read_orbit_property_file(self):
+        """Read the file `*orbclass.out` files
+
+        This file contains angular momenta Lx, Ly, Lz and v_rms of the orbits.
+        Results are stored in and `self.table`
+        
+        """
+        nE = self.settings['nE']
+        nI2 = self.settings['nI2']
+        nI3 = self.settings['nI3']
+        ndither = self.settings['dithering']
+        norb = int(nE*nI2*nI3)
+        ncol = ndither**3
+        nrow = norb
+        orbclass1 = np.genfromtxt(self.mod_dir+'datfil/orblib.dat_orbclass.out')
+        orbclass1 = orbclass1.ravel()
+        orbclass1 = orbclass1.reshape((5, ncol, nrow), order='F')
+        orbclass2 = np.genfromtxt(self.mod_dir+'datfil/orblibbox.dat_orbclass.out')
+        orbclass2 = orbclass2.ravel()
+        orbclass2 = orbclass1.reshape((5, ncol, nrow), order='F')
+        orbclass=np.dstack((orbclass1,orbclass1,orbclass2))
+        orbclass1a=np.copy(orbclass1)
+        orbclass1a[0:3,:,:] *= -1
+        for i in range(int(0), norb):
+            orbclass[:,:,i*2]=orbclass1[:, :, i]
+            orbclass[:,:,i*2 + 1]=orbclass1a[:, :, i]
+        orb_properties = table.QTable()
+        orb_properties['Lx'] = orbclass[0,:,:].T * u.km * u.km/u.s
+        orb_properties['Ly'] = orbclass[1,:,:].T * u.km * u.km/u.s
+        orb_properties['Lz'] = orbclass[2,:,:].T * u.km * u.km/u.s
+        orb_properties['r'] = orbclass[3,:,:].T * u.km
+        orb_properties['Vrms'] = orbclass[4,:,:].T**0.5 * u.km/u.s
+        r_vrms = orb_properties['r']*orb_properties['Vrms']
+        orb_properties['lmd_x'] = orb_properties['Lx']/r_vrms
+        orb_properties['lmd_y'] = orb_properties['Ly']/r_vrms
+        orb_properties['lmd_z'] = orb_properties['Lz']/r_vrms
+        orb_properties['r'] = orb_properties['r'].to(u.kpc)
+        self.table = orb_properties
 
 # end
--- a/dynamite/orblib.py
+++ b/dynamite/orblib.py
@@ -902,18 +902,18 @@ class LegacyOrbitLibrary(OrbitLibrary):
         n_xt_exact = np.sum(bool_xtube)
         n_yt_exact = np.sum(bool_ytube)
         n_zt_exact = np.sum(bool_ztube)
-        def percent(f):
-            return str(int(100*f))
+        def percent(x):
+            return f'{100.*x:.1f}%'
         self.logger.info('Orbit library classification:')
-        self.logger.info(f'    - {percent(n_box/n_orb_tot)}% box')
-        self.logger.info(f'    - {percent(n_xtish/n_orb_tot)}% x-tubes')
-        self.logger.info(f'    - {percent(n_ytish/n_orb_tot)}% y-tubes')
-        self.logger.info(f'    - {percent(n_ztish/n_orb_tot)}% z-tubes')
-        self.logger.info(f'    - {percent(n_other/n_orb_tot)}% other types')
+        self.logger.info(f'    - {percent(n_box/n_orb_tot)} box')
+        self.logger.info(f'    - {percent(n_xtish/n_orb_tot)} x-tubes')
+        self.logger.info(f'    - {percent(n_ytish/n_orb_tot)} y-tubes')
+        self.logger.info(f'    - {percent(n_ztish/n_orb_tot)} z-tubes')
+        self.logger.info(f'    - {percent(n_other/n_orb_tot)} other types')
         self.logger.info('Amongst tubes, % with only one nonzero component of L:')
-        self.logger.info(f'    - {percent(n_xt_exact/n_xtish)}% of x-tubes')
-        self.logger.info(f'    - {percent(n_yt_exact/n_ytish)}% of y-tubes')
-        self.logger.info(f'    - {percent(n_zt_exact/n_ztish)}% of z-tubes')
+        self.logger.info(f'    - {percent(n_xt_exact/n_xtish)} of x-tubes')
+        self.logger.info(f'    - {percent(n_yt_exact/n_ytish)} of y-tubes')
+        self.logger.info(f'    - {percent(n_zt_exact/n_ztish)} of z-tubes')
         self.logger.info('Orbit library classification DONE.')
         # save the output
         orb_classification = {
--- a/dynamite/orblib.py
+++ b/dynamite/orblib.py
@@ -759,14 +759,19 @@ class LegacyOrbitLibrary(OrbitLibrary):
 
     def read_orbit_property_file_base(self, file, ncol, nrow):
         """
-        read in `datfil/orblib.dat_orbclass.out`
+        base method to read in `*orbclass.out` files
 
         which stores the information of all the orbits stored in the orbit library
         norb = nrow = nE * nI2 * nI3 * ndithing^3
         for each orbit, the time averaged values are stored:
-        lx, ly ,lz, r = sum(sqrt( average(r^2) )), Vrms^2 = average(vx^2 + vy^2 + vz^2 + 2vx*vy + 2vxvz + 2vxvy)
+        lx, ly ,lz, r = sum(sqrt( average(r^2) )), 
+        Vrms^2 = average(vx^2 + vy^2 + vz^2 + 2vx*vy + 2vxvz + 2vxvy)
+        The file was stored by the fortran code orblib_f.f90 integrator_find_orbtype.
 
-        The file was stored by the fortran code orblib_f.f90 integrator_find_orbtype
+        *** Do not try to replace this with numpy (e.g. `np.genfromtext`) since
+        the output files are sometimes written in non-standard arrays (in a 
+        seeming system dependent way). **
+
         """
         data=[]
         lines = [line.rstrip('\n').split() for line in open(file)]
@@ -782,8 +787,14 @@ class LegacyOrbitLibrary(OrbitLibrary):
     def read_orbit_property_file(self):
         """Read the file `*orbclass.out` files
 
-        This file contains angular momenta Lx, Ly, Lz and v_rms of the orbits.
-        Results are stored in and `self.table`
+        This file contains time-averaged properties of individual orbits within
+        a bundle. Results are stored in `self.orb_properties`, an astropy table 
+        with columns [r, Vrms, L, Lx, Ly, Lz, lmd_x, lmd_y, lmd_z, lmd] i.e. 
+        the time-averaged value of (i) radius, (ii) V_rms = 
+        (vx^2 + vy^2 + vz^2 + 2(vxvy + vxvz + vyvz)), (iii) total angular 
+        momentum L, (iv) Lx, (v) Ly, (vi) Lz, and circularities (vii) 
+        lmd = L/V_rms, (ix) lmd_x=Lx/V_rms, (x) lmd_y, (xi) lmd_z. Each column
+        of the table is an array of shape [N_bundle, N_orbit].
 
         """
         nE = self.settings['nE']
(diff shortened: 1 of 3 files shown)


dynamite/orblib.py
```diff
@@ -753,4 +755,43 @@ def get_ml_of_original_orblib(self):
...
+        ndither = self.settings['dithering']
+        norb = int(nE*nI2*nI3)
+        ncol = ndither**3
+        nrow = norb
+        orbclass1 = np.genfromtxt(self.mod_dir+'datfil/orblib.dat_orbclass.out')
+        orbclass1 = orbclass1.ravel()
+        orbclass1 = orbclass1.reshape((5, ncol, nrow), order='F')
+        orbclass2 = np.genfromtxt(self.mod_dir+'datfil/orblibbox.dat_orbclass.out')
+        orbclass2 = orbclass2.ravel()
+        orbclass2 = orbclass1.reshape((5, ncol, nrow), order='F')
```
maindlt, line 779: I have not tested the code yet, but shouldn't this line 777 read
`orbclass2 = orbclass2.reshape((5, ncol, nrow), order='F')` (orbclass2.reshape instead of orbclass1.reshape)?

prashjet:
@Steffi-16 - OK, the problem should be fixed, please try again

[The commit message](https://github.com/dynamics-of-stellar-systems/dynamite/pull/285/commits/54f992d572d3467d00568b0f71067033393e5376) explains what was happening. I made a change years ago to read `orbclass.out` files using `np.genfromtxt` rather than line-by-line as had been done in an [older version of dynamite](https://github.com/dynamics-of-stellar-systems/triaxschwarz/blob/sabine/schwpy/schw_AnaResults/schw_orbit.py
). This has (largely!?) not caused any problems. But it turns out that sometimes, `orbclass.out` files are not written out in a regular grid, as expected by `np.genfromtxt`, but in an irregular format. I have no idea why. In any case, in `orblib.read_orbit_property_file` I have reverted to the line-by-line reading.

maindlt:
Hi @prashjet and @Steffi-16,,

I tested this PR using the configuration file `user_test_config_ml.yaml` and existing model data (e.g., the result of running `test_nnls.py`). `orblib.read_orbit_property_file()` and `orblib.classify_orbits()` seem to work as expected. @prashjet: please still validate my fixes, esp. the last commit dfbfb32 (just to make sure).

Below please see the output (copy/paste from a notebook). `orblib.get_projection_tensor()` will get its own post.

Cheers,
Thomas

```
import matplotlib.pyplot as plt
import dynamite as dyn
c = dyn.config_reader.Configuration('user_test_config_ml.yaml',reset_logging=True) # your config file here
mod = c.all_models.get_model_from_row(c.all_models.get_best_n_models_idx(n=1)[0]) # get any model here
orblib = mod.get_orblib()
orblib.read_orbit_property_file()
orblib.orb_properties
```

Lx | Ly | Lz | r | Vrms | lmd_x | lmd_y | lmd_z | L | lmd
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
-1.54006e+18 | 1.38308e+18 | 2.47422e+17 | 0.09822251093826664 | 866.2430374900568 | -0.58659223099282 | 0.5268002433941207 | 0.09424036919126885 | 2.0846852846614523e+18 | 0.7940341233766629
1.54006e+18 | -1.38308e+18 | -2.47422e+17 | 0.09822251093826664 | 866.2430374900568 | 0.58659223099282 | -0.5268002433941207 | -0.09424036919126885 | 2.0846852846614523e+18 | 0.7940341233766629
-1.4357e+18 | 1.34132e+18 | 2.26763e+17 | 0.09877636011883269 | 865.0745632603007 | -0.5445109208542783 | 0.508715879612914 | 0.08600329452230877 | 1.97782589996415e+18 | 0.7501203608545804
1.4357e+18 | -1.34132e+18 | -2.26763e+17 | 0.09877636011883269 | 865.0745632603007 | 0.5445109208542783 | -0.508715879612914 | -0.08600329452230877 | 1.97782589996415e+18 | 0.7501203608545804
-1.06826e+18 | 1.19598e+18 | 1.63302e+17 | 0.10020554378547765 | 853.8424913296362 | -0.4046289098759404 | 0.45300590084195536 | 0.061854520660289465 | 1.611897990321968e+18 | 0.6105447425300932
1.06826e+18 | -1.19598e+18 | -1.63302e+17 | 0.10020554378547765 | 853.8424913296362 | 0.4046289098759404 | -0.45300590084195536 | -0.061854520660289465 | 1.611897990321968e+18 | 0.6105447425300932
-4.02904e+17 | 3.93739e+17 | -9779370000000000.0 | 0.10307622608006747 | 807.9152183242992 | -0.15679302377771534 | 0.15322639732843024 | -0.0038057130059296405 | 5.634338199066479e+17 | 0.21926437146762512
4.02904e+17 | -3.93739e+17 | 9779370000000000.0 | 0.10307622608006747 | 807.9152183242992 | 0.15679302377771534 | -0.15322639732843024 | 0.0038057130059296405 | 5.634338199066479e+17 | 0.21926437146762512
3.86749e+17 | 3.16239e+17 | 9.83046e+17 | 0.09830741935565009 | 710.6335201775947 | 0.17941000961077033 | 0.146700940478968 | 0.4560277914301765 | 1.1027063671884732e+18 | 0.5115373535164711
-3.86749e+17 | -3.16239e+17 | -9.83046e+17 | 0.09830741935565009 | 710.6335201775947 | -0.17941000961077033 | -0.146700940478968 | -0.4560277914301765 | 1.1027063671884732e+18 | 0.5115373535164711
3.28262e+17 | 2.12454e+17 | 9.2714e+17 | 0.09889270409532375 | 715.6521501399965 | 0.15031552334712273 | 0.0972855042532782 | 0.4245497021161492 | 1.006221259147311e+18 | 0.46076205948824067
... | ... | ... | ... | ... | ... | ... | ... | ... | ...
-3650230000000000.0 | 5.77724e+16 | -9535460000000000.0 | 13.985777470354524 | 437.0045766350737 | -1.935517426182259e-05 | 0.0003063354554435527 | -5.056133174255836e-05 | 5.8667703024956584e+16 | 0.0003110827579601535
-1.58368e+16 | 5.13536e+16 | -2.51453e+16 | 14.001495249908329 | 433.27935561251934 | -8.460080056747621e-05 | 0.00027433292533983797 | -0.00013432716903095067 | 5.933196928545352e+16 | 0.0003169536838751691
-4.29924e+16 | 1.8726e+16 | 4.69471e+17 | 14.020518624337367 | 442.16965069981904 | -0.00022474401512362965 | 9.789070689715133e-05 | 0.002454173238156175 | 4.718071951282218e+17 | 0.0024663857657942893
-3.13158e+16 | 6076930000000000.0 | -5.72071e+17 | 14.004444359061724 | 437.0846599916314 | -0.0001657988517327416 | 3.217379137880077e-05 | -0.0030287814748338284 | 5.7295971712578963e+17 | 0.00303348671241266
-3797220000000000.0 | 1.13752e+16 | -3176520000000000.0 | 13.988499724957657 | 438.0114153763575 | -2.0084390031188214e-05 | 6.016610928067696e-05 | -1.6801361686146704e-05 | 1.240581855738669e+16 | 6.56172933258297e-05
-4477120000000000.0 | 2.13959e+16 | -3.91477e+16 | 13.997087790074685 | 437.13727820903125 | -2.371333623770896e-05 | 0.00011332467541821464 | -0.00020734815529468924 | 4.4837167122761e+16 | 0.00023748276122337157
-1.15064e+16 | 1.67665e+16 | -5.55744e+16 | 14.012189821563496 | 442.3991410479907 | -6.0154571166608625e-05 | 8.765396800606128e-05 | -0.0002905386741154118 | 5.91779240812822e+16 | 0.00030937761990193735
-1.9449e+16 | 1.60785e+16 | -5.73763e+16 | 14.0080740318659 | 429.10022139355743 | -0.00010485993017624547 | 8.668776735764115e-05 | -0.00030934622920311135 | 6.268031241897251e+16 | 0.0003379429885175953
-644367000000000.0 | -1832450000000000.0 | -7161120000000000.0 | 13.986360810626623 | 439.9897726084096 | -3.3934029751400597e-06 | -9.65015477483391e-06 | -3.771230667202849e-05 | 7419886891832584.0 | 3.9075039929027134e-05
-580339000000000.0 | 897232000000000.0 | -3.39488e+16 | 14.027421484223884 | 438.73568352710953 | -3.0559792929199683e-06 | 4.72469093572062e-06 | -0.000178769356909464 | 3.3965612611238812e+16 | 0.0001788578896322445
-2153450000000000.0 | 2524040000000000.0 | -2112720000000000.0 | 14.070361809809022 | 439.5418068853064 | -1.1284408333093804e-05 | 1.322635678054382e-05 | -1.1070976885227864e-05 | 3933409541669923.5 | 2.0611669372165955e-05
595291000000000.0 | -1609120000000000.0 | -9754870000000000.0 | 14.039898484488244 | 440.31011798504016 | 3.1207295872636647e-06 | -8.435585946130057e-06 | -5.1138538007311896e-05 | 9904601317871456.0 | 5.1923483443781525e-05

```
orblib.classify_orbits(make_diagnostic_plots=True)
[INFO] 22:54:29 - dynamite.orblib.LegacyOrbitLibrary - Orbit library classification:
[INFO] 22:54:29 - dynamite.orblib.LegacyOrbitLibrary -     - 27% box
[INFO] 22:54:29 - dynamite.orblib.LegacyOrbitLibrary -     - 10% x-tubes
[INFO] 22:54:29 - dynamite.orblib.LegacyOrbitLibrary -     - 7% y-tubes
[INFO] 22:54:29 - dynamite.orblib.LegacyOrbitLibrary -     - 54% z-tubes
[INFO] 22:54:29 - dynamite.orblib.LegacyOrbitLibrary -     - 0% other types
[INFO] 22:54:29 - dynamite.orblib.LegacyOrbitLibrary - Amongst tubes, % with only one nonzero component of L:
[INFO] 22:54:29 - dynamite.orblib.LegacyOrbitLibrary -     - 5% of x-tubes
[INFO] 22:54:29 - dynamite.orblib.LegacyOrbitLibrary -     - 3% of y-tubes
[INFO] 22:54:29 - dynamite.orblib.LegacyOrbitLibrary -     - 23% of z-tubes
[INFO] 22:54:29 - dynamite.orblib.LegacyOrbitLibrary - Orbit library classification DONE.
```

image

`fig = plt.gcf() # save this figure if made successfully`

maindlt:
Hi again,

when trying to run `orblib.get_projection_tensor()`, I get the following error:

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
/var/folders/1b/fphkt4y96ssbkc0ssnj56m1c0000gn/T/ipykernel_19282/1281751111.py in <module>
----> 1 orblib.get_projection_tensor()

~/.local/lib/python3.9/site-packages/dynamite/orblib.py in get_projection_tensor(self, minr, maxr, max_L, nr, nl)
    939         projection = []
    940         for str00 in ['bool_ztish', 'bool_xtish', 'bool_box']:
--> 941             bool00 = self.orb_classification[str00]
    942             # decrease r_idx/L_idx by 1 so they are 0-index
    943             coords = np.array([bundle_idx[bool00], orbit_idx[bool00], r_idx[bool00]-1, L_idx[bool00]-1])

KeyError: 'bool_ztish'
```

This is connected to the classifications the `for` loop expects: While `orblib.get_projection_tensor()` expects `['bool_ztish', 'bool_xtish', 'bool_box']`, the dictionary `orblib.orb_classification` has only the values `{'bool_box':bool_box, 'bool_xtish':bool_xtish, 'bool_ytish':bool_ytish, 'bool_other':bool_other}`, as defined in `classify_orbits()`.

To resolve this we most probably will need to look into why 'bool_ztish' is calculated in `classify_orbits()`, but not put into the `orblib.orb_classification` dictionary and/or `orblib.get_projection_tensor()` uses 'bool_ztish' and 'bool_xtish' but not 'bool_ytish'...

Cheers,
Thomas

Steffi-16:
Hi @prashjet and @maindlt, 
I have tested it again using my config file and models from my run with the BayesLOSVD results, and similarly the `orblib.read_orbit_property_file()` and `orblib.classify_orbits()` work as expected now (I used the same model that showed the error message yesterday). I attach the output of `orblib.classify_orbits()` below. 

```
[INFO] 10:37:12 - dynamite.orblib.LegacyOrbitLibrary - Orbit library classification:
[INFO] 10:37:12 - dynamite.orblib.LegacyOrbitLibrary -     - 43% box
[INFO] 10:37:12 - dynamite.orblib.LegacyOrbitLibrary -     - 5% x-tubes
[INFO] 10:37:12 - dynamite.orblib.LegacyOrbitLibrary -     - 0% y-tubes
[INFO] 10:37:12 - dynamite.orblib.LegacyOrbitLibrary -     - 50% z-tubes
[INFO] 10:37:12 - dynamite.orblib.LegacyOrbitLibrary -     - 0% other types
[INFO] 10:37:12 - dynamite.orblib.LegacyOrbitLibrary - Amongst tubes, % with only one nonzero component of L:
[INFO] 10:37:12 - dynamite.orblib.LegacyOrbitLibrary -     - 77% of x-tubes
[INFO] 10:37:12 - dynamite.orblib.LegacyOrbitLibrary -     - 51% of y-tubes
[INFO] 10:37:12 - dynamite.orblib.LegacyOrbitLibrary -     - 94% of z-tubes
[INFO] 10:37:12 - dynamite.orblib.LegacyOrbitLibrary - Orbit library classification DONE.
```
image

prashjet:
Thanks both for the help so far.  @maindlt - 

> To resolve this we most probably will need to look into why 'bool_ztish' is calculated in classify_orbits(), but not put into the orblib.orb_classification dictionary

This was a mistake on my part. Unfortunately right now I am coding without being able to run dynmaite and debug alongside - hence lots of typos/mistakes. Thank you both for catching them :) I've also checked your last commit Thomas, looks good!

>  and/or orblib.get_projection_tensor() uses 'bool_ztish' and 'bool_xtish' but not 'bool_ytish'...

This was (uncharacteristically) not a error :P In theory there shouldn't be *any* stable y-tube orbits, so I had excluded them from the projection tensor. In retrospect, and seeing in your example Thomas there seems to be 7% of y-tube orbits, maybe this choice was a mistake. So I've added y-tubes back in to the projection tensor.

Steffi-16:
With the newest change, the ` orblib.classify_orbits(make_diagnostic_plots=True)` results in the following error for me: 
image
It works without a problem when setting `orblib.classify_orbits(make_diagnostic_plots=False)`, so I think it is necessary to add `orb_properties = self.orb_properties` in the definition for `find_threshold_angular_momentum` after line 828 (as is done in the definition for `classify_orbits` in line 841)? As that is used when creating the plot and it is not defined there. 

`orblib.get_projection_tensor()` and `print(orblib.projection_tensor.size)` work for me now (when setting `orblib.classify_orbits(make_diagnostic_plots=False)` beforehand) with the following result:
image

prashjet:
> so I think it is necessary to add orb_properties = self.orb_properties in the definition for find_threshold_angular_momentum after line 828 (as is done in the definition for classify_orbits in line 841)? As that is used when creating the plot and it is not defined there.

Thanks - fixed. Also the plotting routine itself is now done - `plotter.orbit_distribution`. Please test :)

Steffi-16:
Using `plotter.orbit_distribution()` with the default arguments, I initially got the error that the argument `orblib` is missing in line 2089 (for the `weights, _, _, _ = weight_solver.solve()`). After adding it, I got the following error for line 2090 concerning `mod_orb_dists = orblib.projection_tensor(weights)`, which I am unsure what to do about: 
image

prashjet:
Please change to

`mod_orb_dists = projection_tensor.dot(weights)`

Steffi-16:
Thank you, that worked! 
However, the resulting plots are all empty:
image

I did not get any error message from this, only the following UserWarnings for lines 2105-2108 in plotter.py:
```
UserWarning: Attempting to set identical low and high ylims makes transformation singular; automatically expanding.
  ax[0].imshow(np.flipud(mod_orb_dists[0]), extent=lmd_rng+log10_r_rng, **kwimshow)
```

prashjet:
It works! But takes horrribly long (and crashes Steffi's computer) for large orbit libraries. Let's fix that. How?

1. double check where the delay is happening! (It might not be where I assumed?)
2. if the delay occurs when getting the weights, then it might be trying to recalculate them, or it is taking longer than expected to read the weights file
3. if the delay occurs at `mod_orb_dists = projection_tensor.dot(weights)` (where I assumed) we can try breaking down the dot product to speed up calculation as follows:

[Whiteboard-[H9VRVPTPR_SbSvYY5F-I-A].pdf](https://github.com/dynamics-of-stellar-systems/dynamite/files/10981086/Whiteboard-.H9VRVPTPR_SbSvYY5F-I-A.pdf)

Steffi-16:
In order to double-check where this delay is coming from, I tested `plotter.orbit_distribution()` again just now and it actually worked a lot faster! After restarting the kernel of the notebook, the small orbit library (6x5x5, dithering=1) was completed within 3 seconds using nr=20, nl=20. 

For the large orbit library (21x10x7, dithering=3) and using the default nr=50, nl=61, it took 1m30s to get the plots (so I think maybe my computer has issues with Zoom rather than the computation itself... I'm very sorry about this!). 

image
image

Is that more like the expected time, or still longer than expected? Comparing this to computing the weights using 
```
orblib = mod.get_orblib()
weight_solver = mod.get_weights(orblib)
weights = weight_solver.solve(orblib)[0]
```
(as is also used in `orbit_distribution()`) the computation time is the same, so I think that this part takes up all the computation time, and that there is no problem with the `mod_orb_dists = projection_tensor.dot(weights)`. 

image

@prashjet, what do you think about this?

prashjet:
Woo - that looks great! That's the basic functionality all done and dusted :)

@maindlt - there are a couple of additional options that would be great to have in the plotting routine. Could you take care of the following:

1. allow an option to plot subsets of orbit classes. I think ideally the method would take a new argument `subsets` e.g. `plotter.orbit_distribution(..., subsets='all')` which could take values `['all', 'short', 'long', 'intermediate', 'box', 'short+box', 'short+long', ...]` etc. Ideally it could accept any possible combination of the orbit classes.
2. allow option for a `orientation=vertical` which would take log r on x-axis and the lambda's on the y-axis

maindlt:
> @maindlt - there are a couple of additional options that would be great to have in the plotting routine. Could you take care of the following:
> 
> 1. allow an option to plot subsets of orbit classes. I think ideally the method would take a new argument `subsets` e.g. `plotter.orbit_distribution(..., subsets='all')` which could take values `['all', 'short', 'long', 'intermediate', 'box', 'short+box', 'short+long', ...]` etc. Ideally it could accept any possible combination of the orbit classes.
> 2. allow option for a `orientation=vertical` which would take log r on x-axis and the lambda's on the y-axis

Sure - will get to it and update tomorrow or Friday (Mar 17) ;-)

maindlt:
Hi @prashjet and @Steffi-16 :-)

The last commits have the plotting enhancements:

- Plotting a subset of orbit classes is done via the new parameter `subsets`. It takes a string and can be 'all' (the default) or one or more of 'long', 'intermediate', 'short', and 'box'. If more than one orbit class is given the classes need to be separated with '+', the sequence does not matter, the plots will always be in the same order. Examples: `plotter.orbit_distribution(..., subsets='all',...)`, `plotter.orbit_distribution(..., subsets='long+short',...)`. To be discussed: when plotting a subset, the total width of the figure stays the same and the panes get expanded. Is this ok or shall we fix the aspect ratio of the individual subplots?

- `orientation='vertical'` is now implemented, in addition to `orientation='horizontal'`. Example: `plotter.orbit_distribution(..., orientation ='vertical',...)`

Please test if you have time ;-)

Thanks and cheers,
Thomas

prashjet:
Apologies: I was not clear what I meant by orientation. I wanted horizontal/vertical to control whether the axes themselves are arranged horizontally or vertically. I also assumed that the choice of what goes on the x or y axis would be fixed by the arrangement: horizontal arrangement would have radius on the y-axis, and vertical arrangement would have radius on the x-axis. This is because all panels share the same radius range, while that isn't true for $\lambda$ (box orbits have $0<\lambda<1$, while others have $-1<\lambda<1$). We can capitalise on this shared r range by passing the relevant `share` option to `plt.subplots`, i.e.

- `fig, ax=plt.subplots(1, n_plots, sharey=True, ...` (horizontal)
- `fig, ax=plt.subplots(n_plots, 1, sharex=True, ...` (vertical)

The benefit of using this `share` option is that we avoid axes labels taking up excess space e.g. in the horizontal case with `sharey=True` only the leftmost axis needs/has a y-axis label. If instead (still for horizontal case) $\lambda was on the y-axis, either (i) each panel would need its own y-axis label (to distinguish the different range for box orbits), or (ii) we still use `sharey=True` but then the box-orbit distribution would only take up half the allowed space. From my perspective, I would prefer to avoid both of these options.

> What do you think of alternative values for the orientation parameter? One way would perhaps be orientation='r_on_xaxis' or orientation='r_on_yaxis'. What do you think?

Given my (longwinded!) though process above, my preference would be to stick to having just a `orientation=horizontal/vertical` option, as defined above. Another possibility would be to have two separate parameters: `orientation = horizontal/vertical` which controls axes arrangement and spearately `r_axis = x/y`. My gut says that's too much flexibility to offer, but happy to hear other opinions.

prashjet:
> To be discussed: when plotting a subset, the total width of the figure stays the same and the panes get expanded. Is this ok or shall we fix the aspect ratio of the individual subplots?

My preference would be to scale  figure size according to number of panels. As before, happy to hear other opinions.

maindlt:
Hi @prashjet and @Steffi-16,

this makes absolute sense, Prash, sorry for the misunderstanding! The last commit (hopefully) implements what you describe:
- `orientation=horizontal` has the subplots share the y-axis (log r)
- `orientation=vertical` has the subplots share the x-axis (log r)
- Both have a color bar which is scaled between 0 and the maximum of the data in the visible frames.
- The figure size scales according to the number of panels (not exactly because of the space for the color bar, this would need a bit more fiddling).
- Changed the dimensions a bit to be consistent with other plots in Plotter.

As usual: any comments are more than welcome! :-)

I also agree to the point that adding the choice between what to plot on the x and y axes would probably be too much ;-)

Please test; below I attach the output of

```
p = dyn.plotter.Plotter(c)
p.orbit_distribution(subset='short+long+intermediate+box')
p.orbit_distribution(orientation='vertical', subset='box+short+long+intermediate')
```

which was obtained with out sample galaxy NGC6278 and the very small orbit library.

Cheers,
Thomas

image
image

dynamite/plotter.py
```diff
@@ -2076,7 +2076,8 @@ def orbit_distribution(self,
                            equal_weighted_orbits=False,
                            orientation='horizontal',
                            figtype='.png',
-                           subset='all'):
+                           subset='all',
+                           getdata='False'):
```
prashjet, line 2080: change string to boolean, i.e. `'False'` --> `False`

dynamite/plotter.py
```diff
@@ -2092,7 +2093,7 @@ def orbit_distribution(self,
             weight_solver = model.get_weights(orblib)
             weights, _, _, _ = weight_solver.solve(orblib)
         mod_orb_dists = orblib.projection_tensor.dot(weights)
-        mod_orbclass_fracs = np.sum(mod_orb_dists, (1,2))
+        mod_orbclass_fracs = np.sum(mod_orb_dists, (1,2))/np.sum(np.sum(mod_orb_dists, (1,2)))
```
prashjet, line 2096: Maybe split into two lines for readability? i.e.

```
mod_orbclass_fracs = np.sum(mod_orb_dists, (1,2))
mod_orbclass_fracs = mod_orbclass_fracs/np.sum(mod_orbclass_fracs)
```

dynamite/plotter.py
Steffi-16: Thank you! Fixed it

prashjet:
Thanks for the changes and testing @Steffi-16! 

We're almost ready to merge. We first just need docstrings for the methods, I'll do by early next week.

dynamite/plotter.py
Steffi-16: Done!

prashjet:
> I assume this is because it only takes a fraction of the full set of orbits in the whole radius range, although I am not entirely sure why because orblib.get_projection_tensor(minr=minr, maxr=maxr, nr=nr, nl=nl) also takes the same radius range.

This is expected. The `projection_tensor` returns total orbit fractions i.e. it doesn't normalise by the radius range. For some other purposes this might be useful, but for these plots it does make sense to normalise.

prashjet:
Doc-strings done. @maindlt please do a read through and merge if happy :)

maindlt:
Looks great, @prashjet and @Steffi-16 :-)

I did some minor edits of the docstrings, made sure the docs compile without errors/warnings, and re-tested the new orbit plots. Will merge now...

PR #263, opened 2022-10-13
New parameter generator SpecificModels generates and runs a predefined list of models

maindlt:
Hi @sthater and @gsantucci90,

Based on our conversation lat week: here is a first version of the new parameter generator `SpecificModels` which generates and runs a predefined list of models. The models can be given as a sequential list or a list of values can be given per parameter and DYNAMITE will build and solve the cartesian product of parameter value combinations.

How to use: see `test_nnls.py`/`user_test_config_fixedvalues.yaml` as an example. The following lines in the config file will create four models with` m in {5.0, 5.5}`,` a in {1e-3, 2e-3}`, and all all parameters at their `value` values. Setting `SpecificModels_mode: "list"` will create two models `m=5.0`, `a=1e-3` and `m=5.5`, `a=2e-3`.

This PR is still a draft so we can fine-tune according to your needs (also needs some more testing...)

Looking forward to discussing in one of our next meetings :-)

```
    bh: # black_hole
        parameters:
            m: # log(BHmass)
                par_generator_settings:
                    lo: 1.0
                    hi: 10.0
                    step: 1.
                    minstep: 0.
                    fixed_values:
                        - 5.0
                        - 5.5
                fixed: True
                value: 5.0
                logarithmic: True
                LaTeX: "$M_{BH}$"
            a: # scale length
                par_generator_settings:
                    fixed_values:
                        - 0.001
                        - 0.002
                fixed: True
                value: 1.e-3  # dot required, renders as str '1e-3' otherwise
                LaTeX: "$a_{BH}$"
...
parameter_space_settings:
    generator_type: "SpecificModels"
    which_chi2: "kinchi2" # "chi2", "kinchi2", or "kinmapchi2"
    generator_settings:
...
        SpecificModels_mode: "cartesian" # "list" or "cartesian"

```
Closes #261.

PR #288, opened 2023-03-29
Tutorial doc fixes: tutorials 1 & 2

### LINKED ISSUE #287: Improve handling of models not created by the model iterator
Apart from using the model iterator, a model can be instantiated directly with a certain parameter set, e.g. by directly calling `model=dyn.model.Model(config=c, parset=parset)`.

The current implementation of this has the following issues which came up when investigating @juliala254's comments in issue #284 (specifically, no. 6). Many thanks, @juliala254!

- If there is no `all_models.ecsv` table, then a standard model directory `orblib_000_000/ml<ml-value of model parset>` will be assigned and subsequently calling `model.get_orblib()` and `model.get_weights()` will calculate the model as illustrated in tutorial `2_quickstart.ipynb`. This works, but some issues arise here:
  - Subsequently changing the parameter set (like suggested in the mentioned tutorial) and re-instantiating the model will lead to the same standard model directory. Hence, `model.get_orblib()` will think the orbit library has already been created and without explicitly deleting the model directory before integrating the orblib DYNAMITE will silently always use the same orblib no matter what underlying parameters have changed.
  - Deleting the standard model directory is a quick fix for this behavior, but the issues of (a) no record of the actually used parset and (b) losing the calculation results of the ‘manually’ created models remain.
  - There is no record of the actual parameter set of this model because there is no `all_models.ecsv` file and none is created.

- If there is an existing all_models table, then trying to directly instantiating a new model via `model=dyn.model.Model(config=c, parset=parset)` will raise an exception if the parset is not existing in the all_models table already.

To my mind a solution to this should make sure that (i) explicitly generated models do not interfere with possibly existing models in the all_models table, (ii) it is possible to run explicit models if an all_models table is existing, and (iii) parset values of those models are recorded.

Ideas:
- Add the explicitly generated models into the all_models table but use a different directory naming scheme (e.g. `orblib_explicit_<index>` instead of `orblib_xxx_yyy`) or to keep them in a separate table like `manual_models.ecsv`. The first alternative would probably be preferable because it makes all models available to possible future iteration runs…
- Quite pragmatic: discourage the use of 'manually' instantiated models and make the model iterator the only supported way of running models.

Looking forward to any comments and cheers,
Thomas

maindlt:
Updated tutorials `1_data_prep_for_gauss_hermites.ipynb` and `2_quickstart.ipynb`.
Biggest changes:
- `2_quickstart.ipynb` now describes the only supported way to create and run models: using the `ModelIterator` class.
- Tutorials don't have the output as part of the distributed notebooks
- The automatically created tutorial output in the docs does no longer have full paths (potentially containing personal information)

Some minor doc tweaks.

Please have a look at the tutorials 1 and (especially) 2...

Closes #287.


azocchi:
Hello @maindlt, I have tested the new versions of the tutorials, and I think that they are ok! However, please notice that the file `NGC6278_config_single.yaml` is not available in the tutorial directory -- I imagine that this is due to the fact that we are waiting for PR #263 to be merged before including this (to be able to use the new parameter generator), but I am writing this down so we don't forget! ;)
Anyway, I committed some changes, mainly fixing some typos, and I have changed the color maps used in the data preparation file, to make them coherent with the ones used in our `Plotter` class. Please notice that this will only work with the most recent version of `pafit` (2.0.8) -- can you please double check this and, in case this is needed, update the requirements accordingly? Thank you! :)

maindlt:
Many thanks for thoroughly going through the tutorials 1 and 2 and for your improvements, @azocchi :-)

The config file was indeed missing in my earlier commit (sorry...) - I added it in the last commit and yes, we need to change the parameter generator in it to `SpecificModels` (and perhaps include more specific exercises working with a few explicit parameter sets at the end) after merging PR #263 ;-)

Also, I double checked the dependency on the `pafit` version: `pafit 2.0.7` results in an error while the tutorial 1 works with `pafit 2.0.8`. I updated the requirements accordingly.

Cheers,
Thomas

prashjet:
Notebooks 1 and 2 run successfully from start to finish.

Happy to merge.

PR #290, opened 2023-04-06
Plot decomposition

azocchi:
As discussed, new plotting function added.


maindlt:
Hi @azocchi,

many thanks for the new and more concise plotting method 👍 I confirm that the component plots before and after the  changes  (*.png and *_ALI.png, respectively) look the same for both our example galaxy NGC6278 and FCC... (both kinematics).

With the plots confirmed to be the same, I'd like to propose the following:

- It might be on my machine only, but I have the impression that the plots look a bit cramped (legends of the color bars overlap with the neighboring plots or are invisible outside the plot area).
- Also, I think the plots should have axis titles on the x and y axes.

If this is fixed (and you agree) I recommend merging this PR into #281.

Side comment: there seems to be a problem with histogram normalization in some bins in `dyn.kinematics.Histogram.get_mean_sigma_gaussfit()`, I will add a comment in the 'main' PR #281 (decomposition1 branch) and deal with it there...

Cheers,
Thomas

azocchi:
Hi @maindlt,

I managed to include the following things in my last commit:
- I (hopefully) made the plot less cramped (with figure size adjusting according to the values of `xlim` and `ylim`), and I changed the color scheme to be the same we are using for the main plots in DYNAMITE
- I removed the function `plot_comps_giu`, and renamed the output plots/files (removing "test" and "_ALI")
- I added a (small) documentation string on the function `plot_comps` (please double check if this is ok)
- I did not add axis titles to the panels in the figure because we also do not have them in the kinematic maps plot (but it is trivial to add them everywhere, and we can easily do it even after merging this branch!)

If you have time, please take another look at this, and then we can merge it!

Thanks! :)

PR #291, opened 2023-04-07
Re-created test galaxy kinematics

maindlt:
In our last hack session we saw that the gh kinematics of our test galaxy NGC6278 has nonzero h3 coefficients which actually should be all zero - a big thank you to @azocchi for noticing this!

This PR re-creates the test galaxy input data in `dev_tests/NGC6278_input/` from our tutorial Voronoi binned kinematics `.fits` file and updates the comparison data for `test_nnls.py`. Now all h3 (and h4, as before) coefficients are zero.

1. Please check/test whether this makes sense.
2. In the tutorials, we have only one mge file (`docs/tutorial_notebooks/NGC6278_input/dynamite_input/mge.ecsv`) which the config file links to both mge_pot and mge_lum. In `dev_tests/NGC6278_input/` however, we have `mge_pot.ecsv` and `mge_lum.ecsv` which differ by one line. Shall we unify this to make `dev_tests/` 100% compatible with the tutorial notebooks? If so, how?


azocchi:
Hi @maindlt, I think that this is all ok: the input files in the `dev_tests/NGC6278_input/` directory seem correct to me now! We should indeed discuss how to keep things consistent between this directory and `docs/tutorial_notebooks/NGC6278_input/dynamite_input/` (also to avoid having the same files twice...).

maindlt:
Many thanks @azocchi, I will go ahead and merge into master...

PR #294, opened 2023-04-20
Clean aperture.dat and source of position angle

maindlt:
In our last hack session we were confused by the last line in `aperture.dat` which is `aperture = -(hst_pa) + 90`.

Checking DYNAMITE's code confirmed that this last line is ignored when reading the file. Reading is performed by subroutine `aperture_boxed_readfile()` in `orblib_f_new_mirror.f90`, subroutine `aperture_boxed_readfile()` in `triaxmassbin.f90`, `Integrated.read_aperture_and_bin_files()` in `data.py`, and `Plotter._plot_kinematic_maps_gaussherm()` in `plotter.py`.

This PR
- removes this last line when writing `aperture.dat` (annotating the primary reading method in `data.py`)
- removes the `position_angle` system attribute from the config file because DYNAMITE does not use it (the position angle is read from `aperture.dat`) along with updating the docs.

So far tested with `test_nnls.py`, let's discuss in the next hack session...


azocchi:
Hi Thomas! Everything looks great to me, you can go ahead and merge this when you want! :) Thanks!

PR #295, opened 2023-05-01
Improved legacy_fortran error messages and handling

### LINKED ISSUE #260: Better manage Fortran errors
If legacy_fortran fails computing an orblib or weight solving (which can happen due to various reasons), DYNAMITE will set `chi2=kinchi2=kinmapchi2=np.nan` and (if applicable)` orblib_done=False `and `weights_done=all_done=False`, log an error but continue running, effectively ignoring the respective models.

As this is confusing to users, we should definitely add an explanation, including one or more of

- change the error messages to warnings and/or
- add text to the log messages that explains that DYNAMITE will continue just ignoring the failed models and/or
- clean up files after a failed orbitstart (and potentially other legacy_fortran programs that may fail).

maindlt:
If legacy_fortran fails computing an orblib or weight solving (which can happen due to various reasons), DYNAMITE will set `chi2=kinchi2=kinmapchi2=np.nan` and (if applicable) `orblib_done=False` and `weights_done=all_done=False`, log an error but the model iterator will continue running, effectively ignoring the respective models.

Some users were confused by the `ERROR` that was recorded in the log in those cases. Therefore, logging will now issue `WARNING`s instead of `ERROR`s because DYNAMITE's model iterator can continue to run when legacy_fortran fails for individual orbit libraries.

However, in some cases the errors were caused by an incomplete DYNAMITE installation or a wrong `legacy_directory` setting so that DYNAMITE did not find the legacy_fortran executables. Such `command not found` errors will now create an error stopping DYNAMITE.

Same behavior as before: If the methods `Model.get_orblib()` or `Model.get_weights()` are called directly for specific models (i.e., not by the model iterator) and DYNAMITE decides to perform orbit integration and/or weight solving, any Fortran error will stop DYNAMITE.

Closes #260.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Compiled source #
 ###################
 .DS_Store
+.venv
 dynamite/__pycache__
 legacy_fortran/galahad-2.3/bin/sys/*
 legacy_fortran/galahad-2.3/makefiles/*
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -15,7 +15,7 @@ Change Log
 - New feature: Added a new method AllModels.make_best_models_table() that creates a table of the best models (best n models or models within a chi2-threshold of the best) and saves it to disk
 - Bugfix: If reattempt_failures=False, in certain cases it could occur that orblibs of successful models were deleted
 - Bugfix: Fixed a bug related to a nonexistent model directory if a crash occurs between the parameter generator adding a model and starting to solve it
-- Improvement: Dynamite will no longer crash upon Legacy Fortran errors, but issue warnings and assign nan to the affected chi2 values
+- Improvement: Dynamite will no longer crash upon Legacy Fortran errors (except when executables are not found), but issue warnings and assign nan to the affected chi2 values
 - Improvement: When executing a dummy run (do_dummy_run==True), model_iterator will set both kinchi2 and kinmapchi2 to nan (instead of zero)
 - Improvement: DYNAMITE will retrofit existing all_models tables with the new column kinmapchi2 and calculate its values for existing models whenever possible
 - New feature: chi2 can now be directly calculated from the kinematic maps when using the LegacyWeightSolver via which_chi2: "kinmapchi2"
--- a/dynamite/model.py
+++ b/dynamite/model.py
@@ -905,7 +905,7 @@ class Model(object):
                     config=self.config,
                     directory_with_ml=self.directory)
         else:
-            raise ValueError('Unknown WeightSolver type')
+            raise ValueError(f'Unknown WeightSolver type {ws_type}.')
         weights, chi2_tot, chi2_kin, chi2_kinmap = weight_solver.solve(orblib)
         self.chi2 = chi2_tot # instrinsic/projected mass + GH coeeficients 1-Ngh
         self.kinchi2 = chi2_kin # GH coeeficients 1-Ngh
--- a/dynamite/orblib.py
+++ b/dynamite/orblib.py
@@ -449,6 +449,11 @@ class LegacyOrbitLibrary(OrbitLibrary):
         cmd_string = 'cmd_tube_box_orbs'
         txt_file = open(cmd_string, "w")
         txt_file.write('#!/bin/bash\n')
+        txt_file.write('# first, check whether executables exist\n')
+        for f_name in orb_prgrm, 'triaxmass', 'triaxmassbin':
+            txt_file.write(f'test -e {self.legacy_directory}/{f_name} || ' +
+                           f'{{ echo "File {self.legacy_directory}/{f_name} ' +
+                           'not found." && exit 127; }\n')
         txt_file.write('(rm -f datfil/orblib.dat.tmp datfil/orblib.dat\n')
         txt_file.write(f'{self.legacy_directory}/{orb_prgrm} < infil/orblib.in '
                         '>> datfil/orblib.log\n')
@@ -485,6 +490,11 @@ class LegacyOrbitLibrary(OrbitLibrary):
         cmdstr_tube = 'cmd_tube_orbs'
         txt_file = open(cmdstr_tube, "w")
         txt_file.write('#!/bin/bash\n')
+        txt_file.write('# first, check whether executables exist\n')
+        for f_name in orb_prgrm, 'triaxmass', 'triaxmassbin':
+            txt_file.write(f'test -e {self.legacy_directory}/{f_name} || ' +
+                           f'{{ echo "File {self.legacy_directory}/{f_name} ' +
+                           'not found." && exit 127; }\n')
         txt_file.write('rm -f datfil/orblib.dat.tmp datfil/orblib.dat '
                        'datfil/orblib.dat.bz2\n')
         txt_file.write(f'{self.legacy_directory}/{orb_prgrm} < infil/orblib.in '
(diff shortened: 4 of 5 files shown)


maindlt:
Ideas for testing:

- Rename or delete legacy_fortran executables in the installation directory `.../site-packages/dynamite...egg/legacy_fortran` and verify that `test_nnls.py` crashes with the right error message (`FileNotFoundError`).
- Provoke a different Fortran error (e.g., by including an `exit 1` or similar in the bash command scripts) and verify the `WARNING` log messages written by `test_nnls.py` which (in that case) will lead to a later crash, as indicated in the respective log message. The bash scripts are written in `orblib.py` (`LegacyOrbitLibrary.write_executable_for_ics()`, `LegacyOrbitLibrary.write_executable_for_integrate_orbits_par()`, `LegacyOrbitLibrary.write_executable_for_integrate_orbits()`) and `weight_solvers.py` (`LegacyWeightSolver.write_executable_for_weight_solver()`), respectively.

prashjet:
@maindlt - I think this is unrelated to the specific errors this PR is about, but I am getting an error when I running the tests e.g.

 ```[ERROR] 14:38:34 - dynamite.orblib.LegacyOrbitLibrary - cmd_orb_start exit code 127. ERROR. Message: cmd_orb_start: line 2: /opt/miniconda3/envs/dcdirect/lib/python3.10/site-packages/legacy_fortran/orbitstart: No such file or directory```

The copy of the`legacy_fortran` directory doesn't even exist:

```
(dcdirect) prash@air > ls /opt/miniconda3/envs/dcdirect/lib/python3.10/site-packages/legacy_fortran
ls: /opt/miniconda3/envs/dcdirect/lib/python3.10/site-packages/legacy_fortran: No such file or directory
```

but the executables are all made correctly in `dynamite/legacy_fortran`.

This is from the master branch, just pulled. Is this the problem you discussed last time...? Would you have expected this to be solved...?

maindlt:
Hi @prashjet, strange. I would have expected this installation issue to be solved. As per commit 1ae1c67 (last change to `setup.py`), `setup.py` should create `legacy_fortran` as a subdirectory inside the `dynamite...egg` installation folder. On my machine (I do not use miniconda environments, which I probably should...) it looks as follows:

```
Thomass-MacBook-Pro:~/.local/lib/python3.9/site-packages$ ls legacy_fortran
ls: legacy_fortran: No such file or directory
Thomass-MacBook-Pro:~/.local/lib/python3.9/site-packages$ ls dynamite-3.0.0-py3.9.egg/
EGG-INFO       dynamite       legacy_fortran
Thomass-MacBook-Pro:~/.local/lib/python3.9/site-packages$ ls dynamite-3.0.0-py3.9.egg/dynamite/
__init__.py          config_reader.py     kinematics.py        myrand.py            physical_system.py
__pycache__          constants.py         mges.py              orbit_exploration.py plotter.py
_version.py          data.py              model.py             orblib.py            populations.py
analysis.py          data_prep            model_iterator.py    parameter_space.py   weight_solvers.py
Thomass-MacBook-Pro:~/.local/lib/python3.9/site-packages$ ls dynamite-3.0.0-py3.9.egg/legacy_fortran/
modelgen          orblib            triaxmass         triaxnnls_CRcut
orbitstart        orblib_new_mirror triaxmassbin      triaxnnls_noCRcut

```

and with

```
legacy_settings:
    directory: "default"
```

in the configuration file dynamite finds the executables. Can you verify that you are using this `"default"` setting? If no, I will try with the miniconda environment on the weekend ;-)

On the bright side: as per this PR, when DYANMITE cannot find an executable an error 127 shall be reported and execution ended. So at least this works as expected...

maindlt:
Sounds good, I'll open another issue based on the error you are getting ;-)

prashjet:
> Ideas for testing:
> 
> * Rename or delete legacy_fortran executables in the installation directory `.../site-packages/dynamite...egg/legacy_fortran` and verify that `test_nnls.py` crashes with the right error message (`FileNotFoundError`).

Confirmed this behaviour does occur.

> * Provoke a different Fortran error (e.g., by including an `exit 1` or similar in the bash command scripts) and verify the `WARNING` log messages written by `test_nnls.py` which (in that case) will lead to a later crash, as indicated in the respective log message. The bash scripts are written in `orblib.py` (`LegacyOrbitLibrary.write_executable_for_ics()`, `LegacyOrbitLibrary.write_executable_for_integrate_orbits_par()`, `LegacyOrbitLibrary.write_executable_for_integrate_orbits()`) and `weight_solvers.py` (`LegacyWeightSolver.write_executable_for_weight_solver()`), respectively.

When I added `exit 1` to the start of the `cmd_orb_start` (which is written by `orblib.write_executable_for_ics`), the expected behaivour occured. The followed warning appears in the output:
```
[WARNING] 16:53:42 - dynamite.orblib.LegacyOrbitLibrary - ...failed! cmd_orb_start exit code 255. Message: cmd_orb_start: line 2: exit: 1/opt/miniconda3/envs/dcdirect/lib/python3.10/site-packages/dynamite-3.0.0-py3.10.egg/legacy_fortran/orbitstart: numeric argument required
```
and later a crash is provoked.

If I added `exit 1`  to the _end_ of `cmd_orb_start` however, no error is provoked.

maindlt:
Many thanks for testing, Prash!

Unfortunately , my "ideas for testing" were imprecise: instead of `exit 1` some output needs to be created (e.g., insert `echo x; exit 1`). Sorry for that.

Background: in order to detect whether a bash script created by DYNAMITE fails, we cannot use the exit status because the legacy_fortran programs do not always exit with a return code != 0 when an error occurs (reason: in different cases the Fortran statements STOP, EXIT with and without arguments are used). However, we know that if the script calling the legacy_fortran executables runs correctly, there will be no output at all. Hence, in `LegacyOrbitLibrary.get_orbit_ics()`, `LegacyOrbitLibrary.get_orbit_library_par()`, `LegacyOrbitLibrary.get_orbit_library()`, and `LegacyWeightSolver.solve()` we check for no output as an indication for a successful run and only if there is output from the script (which needs to be an error message then) we display a warning (or error if the error code is 127).

Producing some output along with the exit should result in a warning/error...

maindlt:
Addendum: to avoid errors like
```
[WARNING] 11:22:04 - dynamite.orblib.LegacyOrbitLibrary - ...failed! cmd_orb_start exit code 1. Message: At line 192 of file orbitstart.f90 (unit = 31)
Fortran runtime error: Cannot open file 'datfil/begin.dat': File exists
```
we could delete `datfil/begin.dat` (if it exists) in the shell script that executes `orbitstart`. In that case, `orbitstart` would just replace the existing file silently. We do the same in the scripts that integrate the tube and box orbits (there, we delete `datfil/orblib.dat` and `datfil/orblibbox.dat` before we integrate).

prashjet:
Sounds like a good idea, but I guess this is quite a contrived situation which is unlikely to come up in real use. Then again, I feel 99% sure that this would be a benign change with low chance for unintended consequences, so why not just do it? Though are these famous last words...?

As you can see, I'm undecided :P What does you think?

maindlt:
Hm, it won't happen often, but on the other hand...

What about this commit: Avoiding the famous last words (and messing with the files from within bash scripts) I now changed `orblib.get_orblib()` so that it checks whether `datfil/begin.dat` and `datfil/beginbox.dat` exist and if both exist `orblib.get_orbit_ics()` is skipped. Otherwise it calls `orblib.get_orbit_ics()` after making sure none of the two files exist.

prashjet:
That commit looks good to me. Much more transparent there than in the bash script.

Go for the merge :)

PR #300, opened 2023-05-12
Make without galahad

maindlt:
DYNAMITE can now be built without GALAHAD (`LegacyWeightSolver` will not be available then, only `NNLS`/`scipy`).
Use `make nogal` instead of `make all`. This should not require GALAHAD any more.

Please test that - after `make distclean` and deleting the entire dynamite (incl. legacy_fortran) installation directories - DYNAMITE installs correctly. Especially, `setup.py` should install the legacy weight solver files only if they are present in the `legacy_fortran` directory.

Connected to this: do we still need `modelgen` in legacy_fortran?

### DIFF
--- a/docs/getting_started/installation.rst
+++ b/docs/getting_started/installation.rst
@@ -125,10 +125,17 @@ Download from `github <https://github.com/dynamics-of-stellar-systems/dynamite>`
 
 If you encouter problems during the installation process, have a look at the section :ref:`troubleshooting`. Some of the most common issues are gathered there.
 
-The installation of DYNAMITE consists of three steps, as detailed below.a
+There are two major installation options for DYNAMITE:
+
+(A) DYNAMITE with the GALAHAD (the ``LegacyWeightSolver``) and the SciPy-based ``NNLS`` weight solvers.
+(B) DYNAMITE with just the SciPy-based ``NNLS`` weight solver.
+
+In case of installation option (A), the installation of DYNAMITE consists of three steps; in case of (B), it consists of two steps, as detailed below.
 
 1. Installation of GALAHAD
 --------------------------
+
+NOTE: this section (Installation of GALAHAD) only applies to DYNAMITE installation option (A).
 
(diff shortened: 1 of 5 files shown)


maindlt:
Here are instructions for testing whether DYNAMITE installs correctly with and without GALAHAD.

Make and install with GALAHAD:

0. Delete the DYNAMITE installation directory. Its path should be something like `~/.local/lib/pythonX.X/site-packages/dynamite...egg/`. If existing, also delete `~/.local/lib/pythonX.X/site-packages/legacy_fortran/`.
1. Make sure GALAHAD is installed, following the known procedure.
2. In `legacy_fortran/`: `make distclean; make all`
Now the executables `orbitstart`, `orblib`, `orblib_new_mirror`, `triaxmass`, `triaxmassbin`, `triaxnnls_CRcut`, `triaxnnls_noCRcut` should be built.
3. After `python setup.py install --user` in the `dynamite/` directory, the DYNAMITE Python files should be installed in the `dynamite...egg/dynamite/` installation directory, along with the aforementioned Fortran executables in `dynamite...egg/legacy_fortran/`.

Make and install without GALAHAD:

0. Delete the DYNAMITE installation directory. Its path should be something like `~/.local/lib/pythonX.X/site-packages/dynamite...egg/`. If existing, also delete `~/.local/lib/pythonX.X/site-packages/legacy_fortran/`.
1. In `legacy_fortran/`: `make distclean; make nogal`
Now the executables `orbitstart`, `orblib`, `orblib_new_mirror`, `triaxmass`, `triaxmassbin` should be built (but NOT `triaxnnls_CRcut`, `triaxnnls_noCRcut`).
3. After `python setup.py install --user` in the `dynamite/` directory, the DYNAMITE Python files should be installed in the `dynamite...egg/dynamite/` installation directory, along with the aforementioned Fortran executables in `dynamite...egg/legacy_fortran/`.

azocchi:
Everything looks great to me, please go ahead and merge when you are ready! :)

PR #304, opened 2023-05-22
Reconstruct 3dmodel

prashjet:
This PR adds support for getting intrinsic model moments. The results should be comparable to existing `nn_intrinsic_moments.dat` files.

@Steffi-16 could you test this? The main method is `orblib.get_model_intrinsic_moment_constructor`. The dosctring for that should explain how to use, but please edit/request clarifications if it is unclear. Briefly, the method returns a function:

```
mod = c.all_models.get_model_from_row(0)
orblib = mod.get_orblib()
mod_intmom_constructor, bin_edges = orblib.get_model_intrinsic_moment_constructor()
```
and this function receives weights and returns intrinsic model moments:
```
weight_solver = mod.get_weights(orblib)
weights, _, _, _ = weight_solver.solve()
intmoms = mod_intmom_constructor(weights)
```


Steffi-16:
I tested the new method on both a Gauss Hermite and a BayesLOSVD model, and it worked without a problem for both!
For the Gauss Hermite model, I compared the result with the corresponding `nn_intrinsic_moments.dat` file and found that the values for the old and new intrinsic moments are comparable (but not exactly the same): For the 0th moment, there is a deviation of at most 4.7e-6%, for the first moments 4.9e-6%, and for the second moments 0.001%.

maindlt:
Tests passed - happy to merge ;-)

PR #305, opened 2023-05-31
PSF weight check

### LINKED ISSUE #302: psf weight
The sum of all elements in the kinematic psf "weight" have to be 1.  We should add a check, that it the case when people run DYNA. Else users might get very wrong results.

maindlt:
Added a check to make sure the sum of all elements in the kinematic psf "weight" = 1.0 (1e-8 accuracy).

So far tested with `test_nnls.py` (only one PSF).

@sthater / @juliala254 : if you have time, please test with your erroneous input data where the PSF weights did not add up to 1.0. DYNAMITE should crash with a useful error message immediately.
If the PSF weights add up to 1.0, DYNAMITE should run as expected.

Is Julia's account still active? I could not add her to the reviewers' list...

Closes #302.

### DIFF
--- a/dynamite/data.py
+++ b/dynamite/data.py
@@ -72,11 +72,21 @@ class Integrated(Data):
         self.aperturefile = aperturefile
         self.binfile = binfile
         super().__init__(**kwargs)
+        self.logger = logging.getLogger(f'{__name__}.{__class__.__name__}')
         if hasattr(self, 'data'):
             self.PSF = self.data.meta['PSF']
+            if abs(sum(self.PSF['weight'])-1.0) > 1e-8:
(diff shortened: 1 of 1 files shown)


juliala2504:
I testet with either only one PSF in a LegacyGridSearch or with two that don't sum up to 1 and receive in both cases the correct error message. Thank you!

Idk why you couldn't add me, probably im not assigned to any project or so, not so familiar with GitHub yet :)

maindlt:
Hi @juliala2504, many thanks for testing! Probably we can merge this PR in tomorrow's hack session :-)

For me GitHub has its mysteries, too ;-) But it worked to type your user id in directly, that's good :-)

Cheers,
Thomas

PR #320, opened 2023-08-29
Re-enable support for directly instantiating a Model object

maindlt:
Re-enable support for directly instantiating a `Model` object (bypassing `ModelIterator`) if the all_models table is empty. Only recommended for testing (e.g., `test_orbit_losvds.py`).

Background:
In general, DYNAMITE only supports models that are created via the `ModelIterator` class, which adds new models to the all_models table (see also PR #288).
The only exception: instantiating a `Model` object directly IF THE ALL_MODELS TABLE IS EMPTY. This may be used for investigating specific orbit libraries without weight solving as it is done e.g., in `test_orbit_losvds.py`.

How to test:
- Run `test_nnls.py` and verify the results look as expected.
- Run `test_orbit_losvds.py` and verify the results look as expected.


prashjet:
> How to test:
> 
> Run test_nnls.py and verify the results look as expected.
> Run test_orbit_losvds.py and verify the results look as expected.

Both work as expected! Will merge :)

prashjet:
Ready to merge after fixing changelog :)

PR #321, opened 2023-08-30
Barcode housekeeping

maindlt:
- Updated `.gitignore` (now includes the bartest_output folder and the new bar legacy_fortran executables).
- Reset CRcut to True for test_nnls.py (to be consistent with the comparison data).
- Avoid the warnings when compiling `legacy_fortran/orblib_f_new_mirror.f90`.

Testing: the only necessary test is to recompile legacy_fortran (`make distclean ; make all`). When doing so, the `Warning: Extension: Unary operator following arithmetic operator (use parentheses) at (1)` messages should not come up any more.


prashjet:
> Testing: the only necessary test is to recompile legacy_fortran (make distclean ; make all). When doing so, the Warning: Extension: Unary operator following arithmetic operator (use parentheses) at (1) messages should not come up any more.

Compiled successfully with no warning. Happy to merge :)

PR #340, opened 2023-10-31
Documentation fix: gh_sys_err

maindlt:
Based on the Fortran code and its interface, `GH_sys_err` has `number_GH` entries. This PR updates the docs in this respect.

Before:
``number_GH``: integer, the number of Gauss-Hermites
    - ``GH_sys_err``: a string of length 2 + ``number_GH`` floats, the systematic error applied to ``V``, ``sigma``, ``h3``, ..., ``hN``
Now:
    - ``number_GH``: integer, the highest order of Gauss-Hermites
    - ``GH_sys_err``: a string of length ``number_GH`` floats, the systematic error applied to ``V``, ``sigma``, ``h3``, ..., ``hN``

Rationale:
1. in `LegacyWeightSolver.create_fortran_input_nnls()`, `number_GH` and the string `GH_sys_err` are written into `nn.in`, the latter with the expl. text `[ systemic error of v, sigma, h3, h4... ]`.
2. In `triaxnnls....f90`'s subroutine `readobservations()`, `number_GH` is read into `hermax` which is used to allocate `kinerrscale(hermax)`, into which in turn `GH_sys_err` is read and applied to the moments. In `triaxnnls_noCRcut.f90` this happens in lines 429-464.


prashjet:
Sorry for hijacking this PR but it seemed like the best place to add this.

Tomorrow let's ask others to prrof-read, then merge both PRs and release :)

PR #344, opened 2023-11-13
Beta plot for python nnls

### LINKED ISSUE #339: Make Plotter.beta_plot() work with Python NNLS
`Plotter.beta_plot()` only works with the `LegacyWeightSolver` (expects the file `nn_intrinsic_moments.out` to exist).

Make it work with all weight solvers.

maindlt:
`Plotter.beta_plot(...)` now uses the existing `LegacyOrbitLibrary.get_model_intrinsic_moment_constructor()` method and does not rely on the `LegacyWeightSolver` output any more.

`get_model_intrinsic_moment_constructor()` calls the rather time-consuming orbit library method `read_orbit_intrinsic_moments(...)` every time it is called. This decreases performance, especially when several models in the beta plot share the same orbit library. Therefore, `read_orbit_intrinsic_moments(...)` can now cache the calculated intrinsic moments in files (switched on by default via the `cache=True` method parameter). In the test script (see below) this reduces the time required for creating the beta plot from 10 seconds to about 1 second.

To test: please run the temporary test script `dev_test/test_beta_plot.py` which works off the existing `user_test_config_ml.yaml` configuration file:
- The weight_solver settings `type: "LegacyWeightSolver"` and `nnls_solver: 1` should produce 4 plots in the plot directory: `newanisotropy_var.png`, `newbetaz_var.png` (both using the new implementation) and `oldanisotropy_var.png`, `oldbetaz_var.png` (using the old implementation, relying on LegacyWeightSolver output). They should be the same.
- The weight_solver settings `type: "NNLS"` and `nnls_solver: 'scipy'` should produce 2 plots in the plot directory: `newanisotropy_var.png`, `newbetaz_var.png` (both using the new implementation) and then crash (because the LegacyWeightSolver output isn't there). These plots should correspond to the ones above.

Before merging:
- Remove the test script `dev_test/test_beta_plot.py`
- Remove the test implementation in `plotter.py` (annotated with `# TODO before merging`).

Closes #339.


prashjet:
I ran the proposed tests and got the expected outcome - great, many thanks Thomas! Will leave the final `TODO`-s and merging to yo.

PR #352, opened 2023-11-28
Sb profile plots

### LINKED ISSUE #351: Create SB profile plots for the kinematic components
Create SB vs. R plots like Fig. 3, top panel in Ding, Y., et al.: A&A 672, A84 (2023), per request of Iris.

maindlt:
The SB profile plots are added to `plotter.py`: `Plotter.plot_sb_profile(...)`.

Tested with `test_nnls.py`, the last 2 lines of that script create the `sb_profile.png` file in the plots directory. Very large scatter in this example (NGC6270), let's discuss... ;-)

Closes #351.

### DIFF
--- a/dev_tests/test_nnls.py
+++ b/dev_tests/test_nnls.py
@@ -133,5 +133,6 @@ def create_comparison_data():
 if __name__ == '__main__':
     # create_comparison_data()
     c = run_user_test()
-
+    p = dyn.plotter.Plotter(c)
+    p.plot_sb_profile(Rmax_arcs=15)
 # end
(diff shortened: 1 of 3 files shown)


maindlt:
The example is not chosen optimally ;-) it is our test galaxy NGC6278 which is non-circular and since I used circular R the scatter is enormous.

While we have the position angle in DYNAMITE, we currently do not have the ellipticity (or a and b of the projected "symmetry ellipse"). One solution would be to leave it to the user to provide an and b (e.g., in arcsecs), another approach to calculate it based on the 2d surface brightness. What do you think?

PR #355, opened 2023-12-04
Use Kinematics.get_data() instead of Kinematics.data in DYNAMITE where possible

### LINKED ISSUE #337: More consistent kinematics data access throughout DYNAMITE
While the `GaussHermite.get_data(...)` method is used throughout DYNAMITE to access GH kinematics data, DYNAMITE still accesses `BayesLOSVD` data via the class attribute `.data` in multiple places.

Change accessing all kinematics data to using the getter method.

maindlt:
This PR minimizes direct access to the `Kinematics.data` attribute and replacing it by calling the `Kinematics.get_data(...)` method.

Rationale:
- For GH kinematics, the attribute `Kinematics.data` contains the input data as read from disk, whereas `Kinematics.get_data(...)` (a) always returns the number of GH coefficients `number_GH` specified in the configuration file and (b) if instructed, applies the systematic errors `GH_sys_err` specified in the configuration file.
- For other kinematics (currently `BayesLOSVD`), `.get_data()` returns a deep copy of the kinematics data.
- For consistency,` Kinematics.get_data()` should be used across DYNAMITE whenever kinematics data is accessed.

Notes:
- If 'dummy kinematics objects' are required, directly accessing the `.data` attribute will still be necessary (such as in `weight_solver.py` lines 189+ and `analysis.py` line 558). Is this reason enough to implement an option to instantiate a Kinematics object with the underlying data passed as an astropy table instead of a filename?
- Implemented a new attribute `Kinematics.n_spatial_bins` that can be used independent of the Kinematics type. For `GaussHermite`, it is the same as `.n_apertures`.

Tested with `test_nnls.py` and `test_notebooks.py`.

Closes #337.

### DIFF
--- a/dynamite/config_reader.py
+++ b/dynamite/config_reader.py
@@ -544,10 +544,10 @@ class Configuration(object):
         for k in stars.kinematic_data:
             if k.type == 'GaussHermite':
                 number_GH = self.settings.weight_solver_settings['number_GH']
-                n_obs += number_GH * len(k.data)
+                n_obs += number_GH * k.n_spatial_bins
             if k.type == 'BayesLOSVD':
-                nvbins = k.data.meta['nvbins']
-                n_obs += nvbins * len(k.data)
+                nvbins = k.get_data().meta['nvbins']
+                n_obs += nvbins * k.n_spatial_bins
         two_n_obs = 2 * n_obs
         return two_n_obs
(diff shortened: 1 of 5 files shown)


prashjet:
I ran `test_notebooks` and all passed

PR #365, opened 2024-02-22
Chi2 plot fixes

### LINKED ISSUE #312: Chi2 map wrong axis label
The Chi2 map plot has wrong axis label, it shows "M-BH" for y label instead of "log(M_BH/M_sun)".

I wanted to change it and first tried "log(M_BH/M_\odot)" since it is a LaTeX description and in latex \odot works but dynamite doesn't take it and gives me error messages back. 

Additionally, if you let two parameters vary the colorbar for the chi2 value is directly on the x axis thus the ticks of the axis and the colorbar overlap.

maindlt:
Chi2 plot:
- Fixed labels (log / non-log and subscript formatting).
- Fixed colorbar overlap with x-axis if only two parameters are varied.
- Added colorbar label $`(\chi^2-\chi^2_\mathrm{min})/\sqrt{2n_\mathrm{obs}}`$
Note: $2n_\mathrm{obs}$ is calculated via `config.get_2n_obs()` which returns the sum of $(2n_\mathrm{obs})$ for all kinematic sets, depending on their type:
for ``GaussHermite``, `n_obs = number_GH * number_spatial_bins`;
for ``BayesLOSVD``, `n_obs = n_LOSVD_bins * number_spatial_bins`.

Closes #312

Tested so far with
- Tutorial notebook 3
- `test_nnls.py` with varying `fixed: False` parameters in its config file `user_test_config_ml.yaml`


boeckera:
I did the tests that Thomas wrote in #380. I honestly do not know if you want to have the plot like this 
kinchi2_plot because I would do several things differently (but plotting is very subjective). For example, if you show data in log(quantity) i would make the scale linear and not log? Also I would always show $10^12$ instead of 1e12. I would also make the limits the same for the same quantities, when they are shown in y-axis and x-axis, so it is easier to compare. Lastly, there are no tick labels for the black-hole mass (x-axis).

maindlt:
Hi @boeckera, many thanks for testing and your comments! Of course it is wrong to show log quantities on a log scale - excellent point (and embarrassing missing it, it's been like that forever...). In the last commit this should be fixed and the axis limits should be consistent. As the plotted quantities should be smaller than about 200 in most instances, I did not touch the $10^\mathrm{X}$ vs. 1eX issue, though...

If you have time, a re-test is highly appreciated, as are any comments :-)

Cheers, Thomas

boeckera:
Hello @maindlt,

I guess it is fine for now? 
kinchi2_plot
Except that we show logarithmic things in log10 not log ;) I changed it haha

I do not see how any of these values should ever accede 200, unless you for whatever reason decide to sample things like black hole mass or dark matter fraction in linear space (would you ever want to do that though? I guess that is a different question). 
I guess the question is also, at least for me, if you want this plot as a quick dirty check of what is going on, or if you want to polish it until it becomes "publishable". I guess that applies to all other plots, too.

maindlt:
Hello @boeckera,
outch, thanks for the log10-fixes!
Well, these graphs are mainly to get a quick overview of the results rather than to produce publishable plots. The ticks are a good example for this: sometimes they are close or even overlapping with neighboring axes and need some tweaking; people may prefer to plot log quantities as such and not on a linear scale, etc. Nevertheless our plots need to be mathematically correct.
A universal solution will need much development time and it is questionable whether we want and have the capacity to go into providing publication-grade plots for every need ;-) In the past the strategy was to provide a minimal set and let users extend/add as they wish. Shall we pick that up in one of our upcoming development meetings...?

boeckera:
Hi @maindlt,

maybe if I do not have anything else to do jajaja.
Sounds great!

Cheers,
Alina

PR #372, opened 2024-03-14
Bin and apertures check

### LINKED ISSUE #370: Bin check
I would suggest to have a double-check of the bins.dat file. The highest number in this file correspond to the total number of kinematic bins. 

Usually, this should be the case. However, Julia and I recently noticed a problem, when the kinematics were derived with IDL. In that bins.dat file the highest number was #bins+1 . It took us 2 hours to find the problem, because DYNAMITE did not provide a good error.

This is just a check to help people finding who create their kinematics with IDL :)

maindlt:
Upon reading the aperture, bins, and kinematics files, DYNAMITE now checks for
- consistent number of velocity (kinematics) bins between the kinematics and bins input files
- consistent number of spatial bins between the aperture and bins input files

Tested with `test_nnls.py` (with provoking errors in all three input files) and "all passed" with `test_notebooks.sh`.

@sthater: perhaps you and/or Julia can test this with your erroneous data and verify that you get meaningful error messages.

Closes #370.

### DIFF
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -7,6 +7,7 @@ Change Log
 Version: 4.2
 ================
 
+- Improvement: improved checks and error messages for velocity and spatial bin input data inconsistencies.
 - Improvement: if ``number_GH`` in the config file is larger than the kinematic order of the observed data, then DYNAMITE ensures that the corresponding systematic errors are > 0.
 - Bugfix: fixed a bug in the kinematics errors (affects NNLS solves).
 - New feature: Gauss-Hermite kinematic maps can now be plotted for any number of Gauss-Hermite coefficients.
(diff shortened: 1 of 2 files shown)


juliala2504:
Tested with a bin.dat that was created with a bintable written in IDL, which starts at 1 instead of 0. Therefore no consistent number of velocity (kinematics) bins between the kinematics and bins input files. 

Old error, that only occured at the end of a model during plotting: 
```
ValueError: All (kin)chi2 values are nan.
```

Now, It successfully raised the following error:

```
ValueError: Numbers of kinematic bins do not match: 149 (length of gauss_hermite_kins_nifs_gh12.ecsv), 149 (last id in gauss_hermite_kins_nifs_gh12.ecsv), max number 150 in bins_nifs_idl.dat.
```

maindlt:
Hello @juliala2504,

many thanks for testing (and sorry for my late reply)! I again successfully ran `test_notebooks.sh` with both Python 3.12 and 3.10 and will merge into master and delete this branch...

Cheers,
Thomas

PR #378, opened 2024-04-13
Legacy nnls file cleanup

### LINKED ISSUE #377: Remove `nn_orbmat.out`
We are currently storing the matrix of size (N_orb, N_bins * 6) in `nn_orbmat.out`. This is >100 larger than any other output file.

In previous versions (schwpy) this file wasn't stored. I decided to store this since it was useful to when re-implementing weight solving in Python. For the dimensions of the problem we were using at the time, `nn_orbmat.out` was no larger than other outputs, so I didn't think it would cause a problem. I didn't correctly extrapolate this change into the future, and more or less forgot about it since then. Clearly this was a mistake - my apologies!

Can we safely delete the `nn_orbmat.out` files? If so, let's do it! If not, then:

1. what data do we need to extract from the files, and 
2. what dynamite methods do we need to modify,

before we _can_ safely delete`nn_orbmat.out`.

maindlt:
Based on today's discussion, this PR implements:
- `LegacyWeightSolver.solve()` deletes the `nn.orbmat.out` file after a successful solve.
- Add LegacyWeightSolver deprecation info to the log file and the changelog in the docs.

Idea for testing: run any set of models with weight solver `type: "LegacyWeightSolver"` and verify that the file `nn.orbmat.out` is not in the model directory.

Closes #377

### DIFF
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -4,6 +4,8 @@
 Change Log
 ****************
 
+- Improvement: ``LegacyWeightSolver`` is now DEPRECATED and will be removed along with GALAHAD in a future version of DYNAMITE. Use weight solver ``type: NNLS`` instead if you can.
+
 Version: 4.2
 ================
 
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -4,7 +4,7 @@
 Change Log
 ****************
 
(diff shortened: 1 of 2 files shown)


prashjet:
Ran `test_notebooks.sh` and all passed successfully. Happy to merge.

Thanks Thomas! :)

PR #382, opened 2024-04-20
Save disk space by cleaning up decompressed files independently from legacy script succeeding or not

maindlt:
Save disk space by cleaning up decompressed files independently from legacy Fortran code succeeding or crashing.
- Moved deleting the decompressed files from the shell script to the method `LegacyWeightSolver.solve()` so deletion takes place independently of any errors that might occur in the shell script and/or the legacy executables called by that script.

Safer removal of `nn_orbmat.out`: check for existing file before deleting.

Suggestion for testing:
1. Provoke an error in `LegacyWeightSolver.solve()`, right after executing the shell script (this simulates a script error). For example, this can be done in `weight_solvers.py` by changing line 316 from `if not p.stdout.decode("UTF-8"):` to `if not p.stdout.decode("UTF-8") and False:`.
2. Run any example using the `LegacyWeightSolver`.
3. Verify that after DYNAMITE crashes, there are no surviving `orblib_*.dat` files in any of the `datfil/` folders (those would be the uncompressed files).
4. Optional: in line 314, change `os.remove(f_name)` to `pass` and re-run. The `orblib_*.dat` files should still be there in the `datfil/` folder(s) after DYNAMITE crashes.

So far, tested successfully with `test_nnls.py` (after changing to the `LegacyWeightSolver` in its config file).

### DIFF
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -4,7 +4,7 @@
 Change Log
 ****************
 
-- Improvement: save disk space by cleaning up decompressed files after a crash and removing unused legacy files nn_orbmat.out after solving.
+- Improvement: save disk space by cleaning up decompressed files after a crash and removing unused legacy file nn_orbmat.out after solving.
 - Improvement: ``LegacyWeightSolver`` is now DEPRECATED and will be removed along with GALAHAD in a future version of DYNAMITE. Use weight solver ``type: "NNLS"`` instead if you can.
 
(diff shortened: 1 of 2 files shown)


prashjet:
I ran the suggested test with `test_nnls` and all `orblib_*.dat` files are removed.

maindlt:
Many thanks, Prash! Will merge and delete branch :-)

PR #383, opened 2024-04-23
New installation procedure

### LINKED ISSUE #381: Make sure installation works with Python 3.12
At least on the Vienna cluster, DYNAMITE does not install when using Python 3.12.

When using the standard installation procedure (python setup.py install --user), apart from numerous deprecation warnings about distutils/numpy.distutils/setup.py, cythonizing fails after throwing Cython errors.

Investigate and fix...

maindlt:
To make DYNAMITE install on Python 3.12, the following changes are applied:
- `numpy` requirements changed from `numpy<1.25,>=1.21` to `numpy>=1.21`. The complicated reason: Older `numpy` versions somehow use a deprecated attribute `ImpImporter` of the module `pkgutil` and that attribute is removed in Python 3.12. The previous limitation `numpy<1.25` had been necessary because of `numba` that was not compatible with newer `numpy` versions (`numba` is an implicit requirement of the `astropy` package which DYNAMITE requires). Current `numba` versions don't have this limitation (tested with Python 3.8...3.12).
- Directly executing `setup.py` has been deprecated for a while. The alternative is to use `pip` which implicitly reads the information in `setup.py`. The new procedure to install DYNAMITE (along with the `legacy_fortran` executables) is via the command `python -m pip install .` or `python -m pip install . --user`. This has been tested with Python 3.8...3.12, is now in the docs, but needs some more testing and the docs need some polishing (the old text is still there, too).
- `setup.py` uses `setuptools`. This is ok, but the environment variable `SETUPTOOLS_USE_DISTUTILS` must not be set (or set to the empty string), otherwise `setuptools` use `distutils` which is deprecated and causes the `pip` installation to crash with Python 3.12. An according note has been added to the docs.

Given the significant changes in how to install DYNAMITE, let's discuss in (one of) the next hack session(s) and test thoroughly...

Closes #381.


prashjet:
Installed in a fresh conda env with python 3.12, and `test_notebooks` passed successfully

boeckera:
Test with python 3.11.5 and notebooks passed, but failed to install dynamite with a fresh conda env python 3.8. It gives me an error that the cvxopt library cannot be build.

boeckera:
it seems to be connected to this https://github.com/cvxopt/cvxopt/issues/78 but no clue how to fix it

afeldmei:
Installation with fresh conda env and python 3.10.14, test_notebooks.sh passed

maindlt:
> it seems to be connected to this [cvxopt/cvxopt#78](https://github.com/cvxopt/cvxopt/issues/78) but no clue how to fix it

Hi @boeckera,

many thanks for testing! The thing is that `pip` does find a binary wheel distribution of `cvxopt` for Python 3.11 on your platform (the` .whl` file), but does _not_ find one for Python 3.8. A list of available platforms is [here](https://pypi.org/project/cvxopt/#files). macOS ARM64 is only supported for Python 3.9 or later. So it downloads the source files (`.tar.gz`) and tries to compile, which fails. On my macOS x86-64 machine a `cvxopt` wheel is available for Python 3.8…3.12, but when I force it to recompile (`python -m pip install . --no-binary cvxopt --force-reinstall`) I get the same error you are getting. From what I see in your linked issue from the `cvxopt` git repo, there seem to be a number of steps necessary to get it compiled from source. Frankly, I do not think that we should go into supporting this, for the following reasons:
- Supporting the building of 3rd party modules uses a lot of resources on our side (see e.g., GALAHAD…)
- `cvxopt` is only one of two options for the Python NNLS solver (the other one being `scipy.optimize`)
- I assume that there is a wheel available for `cvxopt` for any recent system (as it is the case for Python 3.11 on your machine, for Python 3.8…3.12 on my older machine, and 3.10, 3.12 on the cluster)

The question is how we deal with the fact that users may not be able to install DYNAMITE because of `cvxopt`:
1. Just mention a possible issue with building `cvxopt` in the docs, recommending to use a more recent version of Python and linking to the [issue](https://github.com/cvxopt/cvxopt/issues/78) on the `cvxopt` git.
2. Make `cvxopt` an optional requirement in `setup.py` (`extras_require`). However, this would mean that the user has to explicitly include `cvxopt` when installing (via a parameter to `pip`) and the code would need to made aware that the package might be there or not (in the config reader and the weight solver).
3. Require the user to install cvxopt separately (again, the DYNAMITE code would need to be adapted).

I would opt for the first solution, expecting that most users will be able to install DYNAMITE ok (I cannot remember that anyone had problems with `cvxopt` so far), what do you think?

Cheers, Thomas

maindlt:
Per core team meeting:
Go with option 2: Make cvxopt an optional requirement, mention this in the docs

maindlt:
Hi @boeckera, here come some ideas for testing (as discussed today):
- In a clean environment, install DYNAMITE via` python -m pip install .` (executed in the "main" `dynamite/` directory). This does not install `cvxopt` because it is now optional.
  - Try to execute `test_nnls.py` with `type: 'NNLS'` and `nnls_solver: 'cvxopt'` in its config file `user_test_config_ml.yaml`. DYNAMITE should crash and the error message should give a hint on what to do (different solver or install `cvxopt`).
  - Change the weight solver in the config file to `nnls_solver: 'scipy'` and run `test_nnls.py` again. It should finish without an error.
  - Now, install DYNAMITE via` python -m pip install .[cvxopt]` (executed in the "main" `dynamite/` directory). This install includes the optional requirement `cvxopt` and will only succeed if `cvxopt` can be installed on your system "out of the box".
  - Run `test_nnls.py` with both `nnls_solver: 'cvxopt'` and `nnls_solver: 'scipy'` in the config file. Both should complete without an error.
  - Do this for Python 3.12 and at least one earlier version >=3.8.
- Please also have a quick look at the docs, ideally via the "Files changed" tab above to see if the changes make sense ;-)

Many thanks for taking the time for testing :-)

maindlt:
Will merge this PR given the positive comments above by multiple users...

PR #388, opened 2024-05-04
User-defined vmax, smin, smax for kinematic maps

maindlt:
Kinematic map plot: added cbar_lims='user' to the parameters so that user-defined v_max and s_min_max can be passed to the plotting routine.

`Plotter.plot_kinematic_maps()` has a new parameter value `cbar_lims='user'` which allows user-defined limits `v_max=vmax` (the velocity limits are [-vmax,vmax] then) and `s_min_max=[smin,smax]`.

Example use:
`p.plot_kinematic_maps(cbar_lims='user', v_max=200, s_min_max=[0,300])`

Tested with `test_nnls.py` and `test_notebooks.sh` to make sure it doesn't break anything. Also did some plots with custom limits.

@sthater, if you have time, please test, too. Any feedback is highly welcome :-)

Addresses #179.


prashjet (approved):
Changes look good, tests run successfully!

PR #394, opened 2024-06-12
Reduce Python NNLS RAM requirements

### LINKED ISSUE #393: Cannot allocate memory RAM
Dear Dynamite Team,

I am running DYNAMITE models for galaxies observed with the MUSE instrument. I am using DYNAMITE on a cluster with a node of 64 physical cores and 528 GB of RAM. However, I am encountering a Python error that suggests 528 GB of RAM is insufficient. The issue arises with bunzip2 when decompressing the orbit libraries. In the configuration file, I set ncpus = ncpus_weights = "all_available" and also ncpus = ncpus_weights = 32. The code works with orbit library sizes of 6x5x4, but the issue occurs with sizes like 14x7x7. The following is a screenshot of the error.

Captura de pantalla 2024-05-30 a la(s) 09 52 38

Any suggestions to solve this problem will be very helpful.

Thanks for your help and suggestions.

Regards
Erik Aquino.

maindlt:
A user reported DYNAMITE runs out of RAM when there are many concurrent weight solving processes with the Python NNLS weight solver (specific case: 528GB RAM, `ncpus_weights` = 32, orbit library size 14x7x7).

Diagnosis: one of the Python NNLS processes throws a "Cannot allocate memory" error in `LegacyOrbitLibrary.read_orbit_base()` in the line that extracts the orbit library.

Before this PR:
- The unzipped orbit library data is piped into a Python symbol. This data is then written into a binary "FortranFile" which in turn is read by the remaining statements in `LegacyOrbitLibrary.read_orbit_base()`.
- This means that for each of the `ncpus_weights` processes in the model iterator's processing pool, the (potentially very large) orbit libraries are in memory twice.

Solution:
- Instead of unzipping the orbit libraries into a Python symbol, it is directly extracted to disk, creating the FortranFile for further use.
- Still use the command line tool `bunzip2` for performance and of course preserve the zipped version.

Tested:
- test_nnls.py and test_notebooks.sh
- Also, confirmed that the orbit library extracted via `bunzip2 -c datfil/{fileroot}.dat.bz2 > {tmpfname}` is binary equal to the old solution (piping the `bunzip2` output into `aaa.stdout` and have Python write it as a binary file)

@prashjet: It is quite straightforward, but it'll be great if you can have a glance at the code changes (there're not many...).

Closes #393.

### DIFF
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -4,6 +4,7 @@
 Change Log
 ****************
 
+- Improvement: reduced the main memory requirements of the Python NNLS solvers.
 - Improvement: save disk space by cleaning up decompressed files after a crash and removing unused legacy file nn_orbmat.out after solving.
 - Improvement: stability fix in MGE: if q>0.9999 it will be set to 0.9999 (before, it was 0.99999).
 - Bugfix: the chi2 plot now shows correct axis ticks for log quantities.
--- a/dynamite/orblib.py
+++ b/dynamite/orblib.py
@@ -623,16 +623,13 @@ class LegacyOrbitLibrary(OrbitLibrary):
         # unzip orblib to a temproary file with ml value attached
         # ml value is needed to prevent different processes clashing
         ml = self.parset['ml']
-        aaa = subprocess.run(['bunzip2', '-c', f'datfil/{fileroot}.dat.bz2'],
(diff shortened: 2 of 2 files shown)


prashjet (approved):
Code changes look good, and `test_notebooks` passed successfully.

maindlt:
Great, many thanks! :-)

PR #401, opened 2024-07-20
Catch exceptions thrown by Python weight solving

### LINKED ISSUE #400: Weight solving crashes with BayesLOSVD kinematics
When using BayesLOSVD kinematics, DYNAMITE crashes during weight solving for some (not all) models when using the NNLS/scipy weight solver and SciPy versions >= 1.12.0.

Error message: `numpy.linalg.LinAlgError: Matrix is singular.`

Many thanks to @juliala2504 for reporting this!

maindlt:
Proposed solution to issue #400:

This PR catches exceptions thrown by `scipy.optimize.nnls()` and in that case, sets the orbit weights and $\chi^2$ values to nan, does not create the weights file, sets `weights_done = False`, and logs a warning message, hinting towards using `cvxopt` (which does not have the problem). [For completeness, the same logic has been implemented in case `cvxopt` fails, although this does not seem to happen.]
 
- This solution is compatible with the existing implementation using older SciPy versions.
- By setting `weights_done = False`, a subsequent run with `reattempt_failures: True` and another weight solver (`cvxopt` or `scipy<1.12`) will retry weight solving and update the all_models table.
- It is expected that `scipy.optimize.nnls()` will further mature with future SciPy releases, so this solution should (hopefully) be future proof.
- Added information on this behavior in the docs (Overview -> Comments on kinematics).
- Remark: experimenting with the new `atol` argument of  `scipy.optimize.nnls()` did not help (instead of `Matrix is singular`, it produces `Maximum number of iterations reached` exceptions), apart from the code needing to distinguish between SciPy versions (which is a bit clunky).

Tests performed (with Python 3.10 and varying SciPy versions):
- Data provided by @juliala2504 (BayesLOSVD): SciPy 1.11.4 and SciPy 1.14.0 results are the same (> 5 identical digits in `chi2` and `kinchi2`).
- test_nnls.py: SciPy 1.11.4 and SciPy 1.14.0 results are the same (> 6 identical digits in `chi2`, `kinchi2`, and `kinmapchi2`).

If multiple users complain about scipy nnls problems (both `test_nnls.py` and `Tutorial 4. BayesLOSVD and DYNAMITE` run ok with the latest SciPy), we may still restrict `scipy<1.12` until further notice. As this will require periodic checking as SciPy versions keep coming this may not be the best solution, though...

Resolves #400

### DIFF
--- a/docs/getting_started/code_overview.rst
+++ b/docs/getting_started/code_overview.rst
@@ -109,6 +109,7 @@ Comments on kinematics
 ----------------------
 
 LegacyWeightSolver can't be used with BayesLOSVD - use weight-solver type NNLS.
+In some cases, the weight solver ``type: "NNLS"`` ``nnls_solver: "scipy"`` may fail for some models if the SciPy version is >= 1.12. In such cases, it is recommended to use ``type: "NNLS"`` ``nnls_solver: "cvxopt"`` or to reinstall DYNAMITE with ``scipy<1.12`` in ``requirements.txt``.
 
 It is possible to simultaneously fit multiple sets of kinematics in DYNAMITE, which is only supported for Gauss Hermite kinematics. In that case, all input files should be placed in this directory::
 
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -4,6 +4,7 @@
 Change Log
 ****************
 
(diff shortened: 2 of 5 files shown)


prashjet (approved):
tested successfully with `test_notebooks`

Issue #402, opened 2024-07-22
Dynamite produces 'NaN' output for Losvd on modelling larger extent kinematics 


avichaturvedi:
Dear Dynamite team (@maindlt),

I am running dynamite modelling on a galaxy that has kinematics extending out to around 60 arcsec. When I am using the full available kinematics out to 60 arcsec, the dynamite fails  (see the error message from the run at the bottom) whereas, on running models within 25 arcsec, the models runs fine.  

This is exactly what I meant : 
Image1 : Here is the velocity map extending out to around 60 arcsec, In this case, the models try to run the nnls but its produces 'NaN' output in 'nn_nnls.out' and 'nn_kinem.out' file. 

<img width="427" alt="Image1" src="https://github.com/user-attachments/assets/ffd40d68-6e4a-48da-b79e-6ccaa0b958fe">

Image2: Screenshot of attached 'nn_nnls.out' file. 
<img width="565" alt="Image2" src="https://github.com/user-attachments/assets/978158fe-422b-44de-8fa5-73bbf5241e97">

If I use kinematics maps limited to 25 arcsec then the models work fine. 
Image3: 
<img width="427" alt="Image3" src="https://github.com/user-attachments/assets/5aa85902-2d7b-40e6-804f-aabff2209a7a">

Can you please help me to sort out this error? I think it has something to do with the spatial extension of the kinematical map? For your reference here, I have attached the log file as well. 
[dynamite.log](https://github.com/user-attachments/files/16333225/dynamite.log)

Many thanks for your help and suggestions. 

Regards
Avinash

############### Error message from job run #####

The following have been reloaded with a version change:
  1) openmpi4/4.1.1 => openmpi4/4.1.6

[INFO] 11:18:42 - dynamite.config_reader.Configuration - Config file IC4296_const_config_run21100703.yaml read.
[INFO] 11:18:42 - dynamite.config_reader.Configuration - io_settings...
[INFO] 11:18:42 - dynamite.config_reader.Configuration - Output directory tree: Output211007/.
[INFO] 11:18:42 - dynamite.config_reader.Configuration - system_attributes...
[INFO] 11:18:42 - dynamite.config_reader.Configuration - model_components...
[WARNING] 11:18:42 - dynamite.mges.MGE - changing q=1.0 to q=0.9999 for numerical stability.
[WARNING] 11:18:42 - dynamite.mges.MGE - New mge:
     I         sigma      q     PA_twist
[...]

PR #404, opened 2024-07-30
Check for nan values in the kinematics and mges when first reading th…

maindlt:
Check for nan values in the kinematics and mges when first reading the data.

Testing: please place a `nan` in the kinematics and/or mge input file(s) and run any model (e.g., `test_nnls.py`). Upon instantiating the kinematics (mge, resp.) objects, DYNAMITE should abort with an error pointing to the `nan` values.

Created to avoid hard to debug problems like the one reported in issue #402.


maindlt:
Many thanks, @afeldmei, also for identifying the bug earlier! Cheers, Thomas :-)

PR #408, opened 2024-08-04
Remove unused kins and pops weights

### LINKED ISSUE #405: Is the "weight" entry in kinematics and populations configuration actually used?
Investigate and if it isn't used, remove it...

maindlt:
It turns out that the `weight` entry in kinematics and populations configuration is actually _not_ used by DYNAMITE.

- This PR removes the `weight` attribute from kins and pops (this refers not to the weight attribute of the PSFs, just the weight attribute of the kinematics/populations itself!).
- For compatibility with existing config files: If `weight` is given for a kins entry in the config file, a deprecation warning is logged and `weight` is deleted before instantiating the kins objects.
- The `weight` attribute is also removed from the tutorials' config files as well as from the explicit instantiation of `BayesLOSVD` objects in tutorial 4.

Tested so far:
- `test_nnls.py` with line 115 (`weight: 1.0`) commented out: runs successfully.
- `test_nnls.py` with active line 115 (`weight: 1.0`): runs successfully, logs a deprecation warning, and removes the `weights` attribute after reading the kins.
- All tutorial notebooks execute successfully (each notebook executed interactively because batch-executing them gets somehow entangled with my Python environments...).

Closes #405


afeldmei:
I tried that, but there is no file NGC6278_input/gauss_hermite_kins_with_pops.ecsv, only the old gauss_hermite_kins.ecsv

maindlt:
Many thanks for testing, Anja! Did not realize the file was local only (from the coloring efforts). It is now added and testing should work properly... Cheers, Thomas

afeldmei:
There is still "weight" defined in some of the dev_tests/user_test_config*yaml files. In the "'user_test_config_ml.yaml" used by test_nnls.py it is only commented out. Should we remove the "weight" lines entirely, as if it has never existed? Or leave some ghosts of it in no longer used *yaml files, to haunt us and remind us that there once was this thing called "weight"?

maindlt:
This is a good point - I left the "weight" in the dev_test *.yaml files to sort of remind us of the change... But yes, being haunted by reminders of the "weight" thing forever is probably not the best idea...

I'll take the "weight" attribute out of all *.yaml files, test/clean up the *.py files using them, and update this PR later today ;-)

maindlt:
Here is a summary of changes/tests connected to clean up the "weight" attribute in the dev_tests/ folder. Perhaps you can check whether `test_nnls.py` (our most important test script) still runs ok for your setup ;-)

Took "weight" out of:

- `bartest.yaml` (tested `test_bar.py` successfully)
- `dif_dm_halos_config.yaml` (`dif_dm_halos.py` fails, but _after_ successfully assembling the system -> outside the scope of this PR)
- `example_config_for_pops.yaml` (not used anywhere, kept for reference)
- `reimplement_nnls_config1.yaml` and `reimplement_nnls_config2.yaml` (tested `test_reimplement_nnls.py` successfully)
- `test_slurm_config.yaml` (tested `test_slurm.py` successfully)
- `user_test_config.yaml` (tested `test_orbit_losvds.py` successfully)
- `user_test_config_fixedvalues.yaml` (tested with `test_nnls.py` successfully)
- `user_test_config_ml.yaml` (tested `test_nnls.py` successfully)

Also:

- _Deleted_ `test_validate_pqu_config.yaml` and `test_validate_pqu_values.py` because they are outdated (last commit three years ago with an outdated DYNAMITE version).
- _Deleted_ `user_test_config_multi_ml.yaml`, `user_test_config_multi_ml_FCC.yaml`, and `test_multi_nnls.py` because they point to input folder outside of DYNAMITE and cannot be executed by users of the repository.

PR #409, opened 2024-08-12
Bugfix in ParameterGenerator.check_stopping_critera()

maindlt:
Fixed a bug in ParameterGenerator.check_stopping_critera() (that method would never set self.status['stop'] and DYNAMITE would then execute unneeded code).

Tested with `test_nnls.py` (single iteration and multiple iterations) and all tutorial notebooks.


prashjet:
Checked with regular tests and behaviour as expected: stopping criteria now appear in the log files. I found the new documentation a little confusing, so suggested an alternative.

Happy to merge!

maindlt:
Great, many thanks for reviewing and improving the documentation, Prash! :-)

PR #411, opened 2024-09-05
Coloring: phase space binning

maindlt:
Added `coloring.py` holding the class `Coloring` which so far only has one method `bin_phase_space()`. That method does Voronoi binning of the radius-circularity phase space such that each bin contains orbits with a minimum aggregated weight.

Testing: please see the notebook `docs/tutorial_notebooks/coloring.ipynb` (intended for testing purposes only, may become an actual tutorial later). Let's discuss in the upcoming core team meeting...


maindlt:
Add choice of r / log(r). Put this in the projection tensor.

PR #420, opened 2024-11-11
Decomposition: add counter-rotating component

### LINKED ISSUE #419: Decomposition: Add counter-rotating component + there might be a bug for the 'all' component
Could you please add a counter-rotating component in the decomposition class? for the decomp itself and the plotting.
I would highly appreciate it! :)

And there might be some inconsistencies, see the plot below, the 'all' component does not seem like the sum of all the other components. This might be worth a check.

comps_kin_moments_muse

maindlt:
This PR adds the counter-rotating components `cr_thin_d`, `cr_thick_d`, and `cr_disk=cr_thin_d+cr_thick_d` to decomposition.

Also:
- Moved both the component and `ocut` definitions to `Decomposition.__init__()` for consistency. Passing `ocut` to `decompose_orbits()` still works, but causes a deprecation warning.
- Renamed `warm_d` -> `thick_d` for consistency.
- Calculate the total flux via the `all` component rather than adding up individual components' fluxes.

Tested with `dev_tests/test_decomp.py`.

Open:
- Shall we add another component consisting of `disk+cr_disk`?
- Which parts of the hardcoded definitions shall be made accessible via parameters?

@juliala2504, could you have a look and test with your galaxy?

Closes #419.


juliala2504:
Tested with bayes losvd kinematics and an orbit library of nE 7 nI2 5 nI3 5 dithering 3 (dont have results yet for the big orblib)

Plotting: The 'all' component seems very shallow but this might be due to the CR orbits canceling out the non-CR orbits.
I would suggest to implement a check where the weights of the components are summed up in different ways to see whether they're the same eventually e.g. 1) disk + bulge + cr_cold + cr_warm 2) cold + warm + bulge + cr_cold + cold_warm 3) sum up all weights independently of components. 

Counterrotation should mainly be in the center, the velocity plots could mislead here (can this be changed? so that the red and blue shift does not reach to the outskirts since its not physically true?) but in the SB plots it can be seen well. It would be helpful if the colorbar adapts to the values if possible? so we can see some structure in the SB plots although weights are low.

What is strange is that we see the counterrotating nucleus in the think disk and also a bit in the thick disk component. This should not be this way as it should be displayed within „CR thin/thick disk“ only. Possible reason could be that the signs of the losvd histograms for the CR orbits have not been flipped while the lz signs have been?


comps_kin_moments_muse


Calculating kinematics of components (def comps_aphist):
It only calculates the lsb, velocity and sigma for the 'all' component, the other ones are missing, not sure why?

Open questions:
- Disk+cr_disk is not needed 
- Besides the ocut it would be nice to choose the components for the plotting i.e. when you only want to investigate the cold and warm orbits you can choose those and dont get every single component in the plot

maindlt:
> * Besides the ocut it would be nice to choose the components for the plotting i.e. when you only want to investigate the cold and warm orbits you can choose those and dont get every single component in the plot

Implemented in last commit.

maindlt:
> Calculating kinematics of components (def comps_aphist):
> It only calculates the lsb, velocity and sigma for the 'all' component, the other ones are missing, not sure why?

This was a bug (fixed in last commit) - many thanks for catching it!

Also added the `decomp_table=True` option. Whet set, it saves the orbit-component mapping (see `dev_tests/test_decomp.py`).

maindlt:
> I would suggest to implement a check where the weights of the components are summed up in different ways to see whether they're the same eventually e.g. 1) disk + bulge + cr_cold + cr_warm 2) cold + warm + bulge + cr_cold + cold_warm 3) sum up all weights independently of components.

As a step towards this, the last commits implement creating a table that has the aggregate weight assigned to each component (via the parameter `Decomposition(..., comps_weights=True)` when instantiating the Decomposition object). It does not do the different summations but this should be easily done, given that there are only few components.

> Counterrotation should mainly be in the center, the velocity plots could mislead here (can this be changed? so that the red and blue shift does not reach to the outskirts since its not physically true?) but in the SB plots it can be seen well. It would be helpful if the colorbar adapts to the values if possible? so we can see some structure in the SB plots although weights are low.

This is also done in the last commits:  Added the `individual_sb_colorbars` parameter to `plot_decomp()`. If set to True, sb plots will have individual colorbar ranges.

juliala2504:
Tested with a orblib 35 11 11.  

Many thanks for addressing the issues mentioned above. 
The tables are reasonable and the dispersion maps look very nice.

We should consider adding the option for individual colorbars for the „all“ and „bulge“ component. Since the orbits cancel each other out, a different measure is needed for these components. As seen in the plot, the current color scale is very shallow, making it difficult to distinguish features. However, every galaxy is different and inidividual adaption of the code will be necessary.

Regarding the flux for „all“: there seems to be an issue here. Unlike velocity, the flux should sum across all components to yield the total surface brightness. In this case nothing should cancel out. Lets talk about that in our meeting.

dyna_FG_08 11_decomp

maindlt:
> We should consider adding the option for individual colorbars for the „all“ and „bulge“ component.

The last commit extends the individual colorbars functionality and renamed the parameter `individual_colorbars`:

- `plot_decomp(..., individual_colorbars=False)` will result in the same colorbar limits for all components
- `plot_decomp(..., individual_colorbars=True)` will result in individual limits
- Limit to surface brightness, velocity or velocity dispersion by passing a dict (like `plot_decomp(..., individual_colorbars={'sb': True, 'vel': False, 'sig': False})` or `plot_decomp(..., individual_colorbars={'vel': True})`

Together with the comps_plot parameter, this should address most use cases.

Please test ;-)

maindlt:
> Regarding the flux for „all“: there seems to be an issue here. Unlike velocity, the flux should sum across all components to yield the total surface brightness. In this case nothing should cancel out. Lets talk about that in our meeting.

I might miss something but I think it is ok because the flux is not given in magnitudes but just in log(flux) so that smaller numbers correspond to smaller flux. Therefore, the flux of the "all" component doesn't go as far down as for the "cr thin disk" for example. What do you think?

maindlt:
Hi @juliala2504,
as discussed earlier: the decomposition plots now have units for the x and y axes (did not add units to the colorbars so the plots don't look too busy...). Also, `plot_decomp()` now has an optional `dpi=` parameter to adjust the plot's resolution. Please let me know what you think...
Cheers, Thomas

juliala2504:
Thanks for the adjustments. Tested with same data as before and large orblib. Looks great! There will always be some individual adaptation necessary for specific galaxies but overall, i think this provides an excellent foundation for decompositions :)

Flux contributions are also reasonable when selecting individual_colorbars={'sb': False, .. }. 

comps_kin_fit_muse

PR #429, opened 2025-01-16
Better logging messages and more stable error handling

maindlt:
In the recent past, several user problems were difficult to diagnose. One reason was that log entries were difficult to connect to the models/model directories where the errors occurred because in multiprocessing, the models' messages are interlaced in the log.

This PR adds model and directory information to those log entries, saving time and effort when debugging.

Bugfix: in some instances DYNAMITE got stuck in individual model directories after noncritical errors, which led to a crash down the road. This should be fixed in this PR.

Tested successfully with `test_nnls.py` and `test_notebooks.sh`.

### DIFF
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -4,7 +4,7 @@
 Change Log
 ****************
 
-
+- Improvement: better error messages to improve debugging with multiprocessing
 - Improvement: minor improvements of tutorials 2 and 3
 
 Version: 4.3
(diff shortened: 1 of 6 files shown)


juliala2504:
Tested successfully with` test_nnls.py `.

Exemplary Output:

[INFO] 12:45:17 - dynamite.model_iterator.SplitModelIterator - ... running model 1 out of 3: NGC6278_output/models/orblib_000_000/ml05.00/.
[INFO] 12:45:17 - dynamite.orblib.LegacyOrbitLibrary - Calculating initial conditions for NGC6278_output/models/orblib_000_000/.

maindlt:
Thanks for testing, @juliala2504! Will merge and delete branch...

PR #431, opened 2025-01-24
Two step coloring prep

maindlt:
As discussed in the last core team meetings, spatial binning for the population data (coloring) will be done ouside the legacy Fortran code.

Towards preparing DYNAMITE for coloring, this PR removes code that passed population data back and forth between Python and Fortran. Also, references to histogram objects were removed from population data.

Testing: beyond just reading dummy population data there is no functionality in place yet. Hence, the goal is to keep a functional master branch with DYNAMITE performing as it did before merging this branch.

Tested successfully with `test_nnls.py` and `test_notebooks.sh` with and without dummy population data and both weight solvers.


maindlt:
> * `read_losvd_histograms()` duplicates the tube orbits with flipped velocities. Applied to the populations’ orbit density, this would mean duplicating the individual densities, effectively multiplying the weight of the tubes by two. Do we need to do this?

This has been implemented in the last commits in order to be compatible with the shape of the model weights.

maindlt:
The last commit splits reading the orblib & histograms via `read_losvd_histograms(pops=False)` (with the old and new orblib file structures) and reading the populations' orbit densities via `read_losvd_histograms(self, pops=True)`.

Reason: while the orblib & histograms are used for weight solving, the populations' orbit densities are used in a second step.
Benefit: only the necessary files (`orblib_qgrid.dat`, `orblib_losvd_hist.dat`, `orblib_pop.dat`) are unzipped and read, saving time and disk space.

prashjet:
I ran the notebook `orbit_densitites.ipynb` successfully. I also checked that things work as expected when manipulating the bins for the standalone population data so that it isn't binned identically to the kinematics - everything looked as expected.

Happy to merge!

maindlt:
Great - many thanks, Prash :-)

PR #443, opened 2025-03-06
Ocut in orbit plot

maindlt:
Improvements to orbit plotting and Decomposition:
- There is a new parameter to `orbit_plot(..., ocut=(0.8, 0.25, -0.25, -0.8),...)`. `ocut` works the same as in the `Decomposition` class and allows to manually set the orbit cuts in $\lambda_z$. The plot will include horizontal dashed lines at each $\lambda_z$ value. While the user can pass any number of cuts, the default has four levels (0.8, 0.25, -0.25, -0.8), interpreted as (lim_cold, lim_warm, lim_hot, lim_cr_warm), decomposing the plot into five regions.
- Added a label explaining the colorbar in `orbit_plot(...)`.
- Minor improvement to `Decomposition`: made the use of `>` and `>=` consistent. Now the regions are consistently defined by `upper_limit >= lambda_z > lower_limit`.

@liquidreams, if you get a chance please test to verify the orbit plot does what you want ;-) Many thanks!

### DIFF
--- a/docs/more_info/changelog.rst
+++ b/docs/more_info/changelog.rst
@@ -4,6 +4,7 @@
 Change Log
 ****************
 
+- Improvement: the orbit_plot now allows custom orbit cuts via the ocut parameter (same behavior as in the Decomposition class)
 - Improvement: added counter-rotating components to decomposition.
 - Bugfix: fix a bug that prevents Bayes LOSVD kinemtic maps from being created more than once
 - New feature: a dark halo is no longer mandatory (models can consist of either zero or one dark halo component)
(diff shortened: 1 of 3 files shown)


maindlt:
Hi Anja,
many thanks for testing! The additional line at -0.8 when using the default `ocut` vector was introduced in order to use exactly the same circularity cuts as in the Decomposition class. Do you think this is confusing? Then we can just delete the lambda=-0.8 from the default ocut vector to keep the exact old behavior as the default.
Cheers, Thomas

PR #446, opened 2025-03-18
Support externally calculated chi2 values for components not interacting with DYNAMITE components

maindlt:
Per Meng's request, details pending...

### DIFF
--- a/docs/getting_started/configuration.rst
+++ b/docs/getting_started/configuration.rst
@@ -102,6 +102,28 @@ The system consists of a number of physical components - e.g. the stars, black h
         - ``logarithmic``: Boolean, whether logarithmic steps should be used for parameter search. If true, then (``value``, ``lo``, ``hi``) must all have log units.
         - ``LaTeX``: LaTeX string for this parameter to be used for plots.
 
+There is a special component ``type: Chi2Ext`` that allows calculating a :math:`\chi^2` externally which will be added to all three :math:`\chi^2` values calculated by weight solving in DYNAMITE (see ``which_chi2`` in `parameter_space_settings`_).
+This may be useful for e.g., gas kinematics that do not affect the potential. That component type has different settings from the ones built in DYNAMITE:
+
+- ``component name``: a descriptive name, but preferably short as this will be used to refer to the component in the code (e.g. ``gas``)
+    - ``type``: ``Chi2Ext``
+    - ``contributes_to_potential``: optional, will always be set to False (not currently used)
+    - ``include``: Boolean, whether to include this component or not. If False, equivalent to omitting this component entirely
+    - ``parameters``: Choose any parameter names as you see fit (e.g., par1, par2,...). Each parameter must have values specified for
+        - ``fixed``: Boolean, whether the parameter is to be kept fixed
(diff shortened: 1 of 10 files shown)


momomaycry:
Wow, it looks nice. Thank you. I will test on real gas data soon.

docs/getting_started/configuration.rst
```diff
@@ -102,6 +102,28 @@ The system consists of a number of physical components - e.g. the stars, black h
         - ``logarithmic``: Boolean, whether logarithmic steps should be used for parameter search. If true, then (``value``, ``lo``, ``hi``) must all have log units.
         - ``LaTeX``: LaTeX string for this parameter to be used for plots.
 
+There is a special component type ``Chi2Ext`` that allows calculating a :math:`\chi^2` externally. After weight solving (resulting in DYNAMITE's three :math:`\chi^2` values, cf. ``which_chi2`` in `parameter_space_settings`_), it will be added to those three :math:`\chi^2`.
```
prashjet, line 105: There is a special component type ``Chi2Ext`` that allows external modules to calculate a :math:`\chi^2` value for ``DYNMAITE`` models. These external :math:`\chi^2` values are calculated after ``DYNMAITE``'s internal :math:`\chi^2` calculations then appended to ``DYNMAITE``'s `all_models`, then can be used to inform the parameter search for later model iterations. The ``Chi2Ext`` component has different settings from the ones built in DYNAMITE:

prashjet:
I've suggested a little rewording, but generally I think it all makes sense!

@momomaycry - please let us know if the current documentation is clear enough to implement the gas model. If not, perhaps we can add a simple toy example of an external module in the _Tutorials_ section.

docs/getting_started/configuration.rst
maindlt: Great suggestion, just pushed the update. Many thanks, Prash :-)

momomaycry:
> I've suggested a little rewording, but generally I think it all makes sense!
> 
> @momomaycry - please let us know if the current documentation is clear enough to implement the gas model. If not, perhaps we can add a simple toy example of an external module in the _Tutorials_ section.

It seems to me that it is clear enough to implement a gas disc aligned with the stellar disc. But I am not sure if other people can understand it well.  I will suggest a toy model included in the tutorials.

sthater:
Huhu,

After some hickups with the installation, I successfully tested all dev tests (also somtimes with different inputs):

*test_nnls.py with Python NNLS-scipy weight solver and legacy weight solver
*dif_dm_halos.py with Python NNLS-scipy weight solver and legacy weight solver
* test_bar.py with Python NNLS-scipy weight solver 
*test_decomp.py with ml-yaml and ml_gas.yaml
—> I did get a warning: /Users/sabine/.local/lib/python3.12/site-packages/dynamite/analysis.py:371: RuntimeWarning: divide by zero encountered in log10
  np.log10(t[c_idx][s])-max_flux[comp],

*test_orbit_losvds.py
*test_reimplement_nnls.py


I also successfully tested the branch with some small ATLAS3D and MAGPI runs :) The bigger data runs still need to be tested more and I will work on that in the new year.

Issue #449, opened 2025-04-08
Color mapping issues in plotter.py (Line 259)


yu-2410:
# File: `dynamite/plotter.py `
Original Code: (Line 259 in `make_chi2_plot` function)
```python
color = colormap(val['chi2t'][k]/chlim)
```
# Problem Description
When `val['chi2t'][k] > chlim`, the input to `colormap` exceeds `1.0`, causing undefined behavior (see [matplotlib documentation](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.Colormap.html#matplotlib.colors.Colormap): For floats, the input must be in the interval `[0.0, 1.0]`).
# Fix
```python
from matplotlib.colors import Normalize
norm = Normalize(vmin=0, vmax=3)  
color = colormap(norm(val['chi2t'][k]/chlim))
```

prashjet:
Hi @yu-2410 - thanks for raising this issue! Your suggested fix looks good to me. Would you be happy and able to put this fix in a pull request?

yu-2410:
Hi @prashjet, I've submitted the PR here: #451.

PR #451, opened 2025-04-20
fix: normalize chi2 value for colormap to avoid exceeding 1.0

yu-2410:
- Related to issue #449 .


prashjet:
Many thanks @yu-2410 - happy to merge :)

PR #455, opened 2025-05-16
Always generate models on restart

maindlt:
Per user request by @liquidreams:

After DYNAMITE converged it may be desirable to add more models to verify the optimal parameter set. This PR makes sure, that upon restart, DYNAMITE will always create models for the first iteration, independent of the stopping criteria.

Also, the following bugfix is included in this PR:
The `SpecificModels` parameter generator no longer crashes when the varied parameters' indices don't start with 0 and have no gaps.

Tested with Iris' test galaxy run.

@liquidreams, please test if this works for you.
Feel fee to try the `SpecificModels` parameter generator to investigate models around the optimum. An example config file is `dev_tests/user_test_config_fixedvalues.yaml`

### DIFF
--- a/dev_tests/user_test_config_fixedvalues.yaml
+++ b/dev_tests/user_test_config_fixedvalues.yaml
@@ -31,10 +31,6 @@ system_components:
                 logarithmic: True
                 LaTeX: "$\\log(M_\\mathrm{BH}/M_\\odot)$"
             a: # scale length
-                par_generator_settings:
-                    fixed_values:
-                        - 0.001
-                        - 0.002
                 fixed: True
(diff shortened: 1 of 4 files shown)


maindlt:
Will close this PR without merging for the following reasons:
- In the motivating example for this PR, the `min_delta_chi2...` criterion was the cause for DYNAMITE not calculating more models because the parameter search had already converged mathematically.
  **The recommended workaround when more models are desired is to set the `min_delta_chi2...` criterion in the config file to a negative value. Many thanks @afeldmei for pointing this out!**
- The mentioned bugfix in the SpecificModels parameter generator is now addressed in #457 which also includes a significant update in the documentation and elaborates on testing suggestions.

PR #472, opened 2025-10-08
Python NNLS fix

maindlt:
Users reported weight solves that fail with Python NNLS (both scipy and cvxopt), but work with the deprecated LegacyWeightSolver. At least some of these problems can be solved for `scipy.optimize.nnls()` by increasing the `maxiter` parameter.

Therefore, the `weight_solver_settings` parameter `maxiter_factor` was added to the configuration file. Its default value is 3, which translates to `maxiter = 3 * n_orbits`, the default value `scipy.optimize.nnls()` uses.

In addition, investigating this issue shed some light on the volatile history of `scipy.optimize.nnls()`:
- It worked well until SciPy 1.11 (based on Fortran)
- Re-implementing it in Python in SciPy 1.12 proved to be unstable and much slower (a factor of more than 18 according to own tests) EDIT: The factor of 18 resulted from a run consisting of 56 Solves in chunks of 8, which was scheduled unfavorably. A better performance estimate is running all 56 in parallel, which resulted in a slowdown-factor of 4.4 - better, but still not good...
- Re-implementing in Cython in SciPy 1.15 resulted in better performance (still slower than SciPy 1.11) and stability
- Re-implementing in C in SciPy 1.16 resulted in comparable performance and stability to SciPy 1.11
Therefore, the requirements for Scipy have been changed to `scipy>=1.11,!=1.12.*,!=1.13.*,!=1.14.*,!=1.15.*` to warrant stable Python NNLS while allowing newer SciPy releases.

Tests so far:
- Successfully ran `test_nnls.py` with different values and with missing `maxiter_factor`
- Successfully ran `test_notebooks.sh`
- Successfully ran suite of 56 problematic weight solves (thanks @afeldmei!): before, `scipy.optimize.nnls()` could solve 9 models and failed  solving 47. With `maxiter_factor=10`, al 56 solved successfully. Tested with SciPy 1.11.4 and 1.16.2.

Please test with any 'problematic' models...

### DIFF
--- a/dev_tests/bartest.yaml
+++ b/dev_tests/bartest.yaml
@@ -153,7 +153,7 @@ system_components:
                 type: GaussHermite # specifies which object class to create
                 hist_width: '1578.6020507812'
                 hist_center: '0.0000'
-                hist_bins: '476'
+                hist_bins: '477'
                 datafile: "kinematics.ecsv"  # both discr. & integrated
                 aperturefile: "aperture.dat" # integrated only
                 binfile: "bins.dat"          # integrated only
@@ -206,6 +206,7 @@ weight_solver_settings:
     # nnls_solver: 1
     type: "NNLS"
     nnls_solver: 'scipy'
+    maxiter_factor: 3  # nnls_solver 'scipy' only; default: 3
     CRcut: True
     regularisation: 0
     number_GH: 4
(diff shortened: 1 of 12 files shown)


afeldmei:
So far, maxiter=10 solved all my previously failed models. I can merge, or should I wait for the two other PRs discussed earlier today?

maindlt:
Great news, thanks for testing, Anja!
The changes should not interfere with the two other PRs but I would still wait until #446 is merged because we are in the middle of final testing that one.
In the meantime I'll also add the `maxiter_factor: 3` default entry to all the other config files to make users aware of the new option.
Cheers, Thomas

PR #485, opened 2025-11-26
Remove unused mandatory component attribute 'contributes_to_potential'

### LINKED ISSUE #484: Remove unused mandatory component attribute `contributes_to_potential`
In the configuration file, we currently have two similar component attributes `include` and `contributes_to_potential`, e.g.:

```
stars:
    ...
    include: True
    contributes_to_potential: True
    ...
```

While `include` has an effect (omitting the component entirely if `include=False`), `contributes_to_potential` is a mandatory attribute but has no effect whatsoever. The latter is also - confusingly - somewhat similar to the planned 'external chi2' feature (see PR #446) which also adds a 'component' that does not contribute to the potential, but has nothing to do with the `contributes_to_potential` attribute.

Therefore, I propose to remove `contributes_to_potential` from DYNAMITE.

maindlt:
Removed
- unused mandatory component attribute `contributes_to_potential`. If it is still in the configuration file, a deprecation warning will be logged.
- unused attribute `System.n_pot` (it was set using `contributes_to_potential` but not used anywhere in DYNAMITE).

Tested new functionality:
- `test_nnls.py` ran as expected, no warnings.
- In `user_test_config_ml.yaml`, added `contributes_to_potential: <any value>` to the attributes of any component and executed `test_nnls.py` again. It still succeeded, but showed a deprecation warning in the log.

Tested existing functionality: executed `test_bar.py` and `test_notebooks.sh` successfully.

Closes #484

Volunteers for testing welcome :-)

### DIFF
--- a/dynamite/plotter.py
+++ b/dynamite/plotter.py
@@ -2468,7 +2468,8 @@ class Plotter():
         _ = model.get_weights(orblib)
         intmoms, int_grid = orblib.read_orbit_intrinsic_moments()
         density, x, y, z = np.moveaxis(intmoms[..., 0:4], -1, 0)
-        weights = model.weights[:, *([np.newaxis] * 3)]
+        # weights = model.weights[:, *([np.newaxis] * 3)]  # Python>=3.11 only
+        weights = model.weights[:, np.newaxis, np.newaxis, np.newaxis]
         r_max = np.sqrt(np.max(x**2 + y**2 + z**2,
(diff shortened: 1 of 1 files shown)


maindlt:
Many thanks for testing, especially catching the 3.10 incompatibility, Prash!

I fixed it in this PR to be able to merge, but let's take the discussion to the new Issue #492 :-)

PR #496, opened 2025-12-17
Pass configuration to external chi2 calculation

### LINKED ISSUE #493: Suggestion: passing config into Chi2Ext when gas and stars share the same potential?
Hi, I am porting a gas model for testing, I noticed that Chi2Ext currently only receives `parset`.

If gas and stars are assumed to share the same gravitational potential, the potential definition depends not only on parset but also on stellar mges and DM halo profile types stored in config.

Would it make sense to also pass config into Chi2Ext, or pass all components contributing to the potential,  to ensure a consistent potential definition between the gas and stellar components and to simplify future joint modelling?

maindlt:
Hi @momomaycry,

Many thanks for testing and your comment. Yes, it makes absolutely sense to pass `config` into `Chi2Ext`.

Adding `config` as a `Chi2Ext` attribute might cause problems however because the `Chi2Ext` object is part of `config` itself, in a recursion-like fashion...

While your approach adding the system component types to `Chi2Ext` may work well, I would propose an even simpler solution in this PR: why not pass the configuration to the method that calculates the external chi2? This will have a minimal performance footprint as it is only one reference passed per model.

What do you think? Please let me know whether this solves your issue or if I missed something :-)

Tested:
- `test_nnls.py` (original version)
- `test_nnls.py` and configuration file `user_test_config_ml_gas.yaml`: this adds a debug message to the log listing the parset (as before) and the system name pulled from the configuration (new).

Cheers and all the best,
Thomas

Closes #493


maindlt:
Many thanks for your answer!

Hm, wouldn't it work to access the system via `config.system` because the system is actually part of the configuration? I suggested passing the complete configuration in case data other than the system data is needed in the external chi2 calculation. We can also pass the system only, which would mean passing `config.system`.

As an example for how the external chi2 method would access the system via `config`, the attached notebook demonstrates how to extract information about the black hole class, dark halo class, and the stellar MGE from the `config` object (without re-parsing any input).

Would that work for you or should we better only pass `config.system`?

> I will also need to add a function to calculate the accelaration for the DM halo, BH etc.

This would ideally be a function/method of the `Chi2Ext` class which is called by that class' `chi2()` method with parset and config as arguments.

> At the same time, I am trying to get the gNFW halo to work in fortran, and writing the function needed to transfer Mvir, c to rho, rs. I can share my codes with you when I finish this part.

That would be great! :-)

maindlt:
Hi Meng,

many thanks for your message. I'll be happy to go over your code and align it with the pull request. Unfortunately, your email attachment got lost (I think it is a GitHub issue), so when you are back, please upload it here at github.com or send it directly to thomas.maindl@univie.ac.at.

In the meanwhile, enjoy your holidays and Happy Chinese New Year!

Kind regards,
Thomas

momomaycry:
> > Hi Thomas,
> > I have gone through the changes and they look all well. But I am a little confused now. Are these changes merged to the main branch or another branch?
> > Best, Meng
> 
> Hi Meng,
> 
> many thanks for checking!
> 
> I merged this entire branch with all the changes into the main branch, so that for tests you can just use the main branch. Another branch (and pull request) will only be needed if we apply any future modifications.
> 
> All the best, Thomas

I see. Thank you! I will continue with this latest version. 

Best,
Meng

PR #513, opened 2026-03-12
Always consider GH_sys_err if specified in the configuration

maindlt:
This PR ensures that Gauss Hermite kinematics systematic errors are always and consistently considered by DYNAMITE:
- If systematic errors `GH_sys_err` are given in the configuration file, then `GH_sys_err` is added to the 'regular' observation errors in quadrature right after reading the input data.
- This changes the previous behavior where the systematic errors were considered in some instances and not in others (e.g., passing to CR_cut, plotting kinematic maps, etc.).

Tests:
- Successfully ran `test_notebooks.sh`
- Executed `test_nnls.py` with the following systematic errors in the configuration file:
`GH_sys_err: '0.0 0.0 0.0 0.0 0.3 0.3 0.6 0.6'` and `GH_sys_err: '0.0 100.0 0.0 0.0 0.3 0.3 0.6 0.6'` and
  - observed the effect on the kinematic maps (sigma residuals are all close to zero in the latter case).
  - In the latter case, `dynamite.config_reader.Configuration` reports the increased `dsigma` value in its DEBUG log entries.
- Ran `test_nnls.py` successfully with weight solver type `LegacyWeightSolver`.

Ready to merge after independent test ;-)


prashjet:
Tested successfully `test_notebooks` and `test_nnls`. Kinematic maps change as expected when supplied with large systematic errors. Happy to merge!

PR #515, opened 2026-03-26
New projected mass calculations

### LINKED ISSUE #511: Don't re-calculate observed projected masses for each orbit library
The legacy Fortran program `triaxmassbin` calculates the projected masses. The issue is that the calculation is done for each orbit library while it depends on observed data only.

- To estimate whether this is significant, `triaxmassbin` has been timed by modifying the Fortran call,
`/usr/bin/time -ao file.time {self.legacy_directory}/triaxmassbin  < infil/triaxmassbin.in >> datfil/triaxmassbin.log`
and setting `ncpus: 1`, `ncpus_weights: 1`, `orblibs_in_parallel: False` in the configuration file.
- The runtimes increase with complexity and can get significant: 
  - `test_nnls.py` with 3016 pixels, 152 spatial (Voronoi) bins, len(mge)=6:
    about **0.36 seconds per orbit library**
  - FCC047 with 2 kinematic sets and a total of 175562 pixels, 837 spatial (Voronoi) bins, len(mge)=12:
    about **46 seconds per orbit library** (tested multiple times...)

--> Investigate how to most efficiently do these calculations only once and implement, ideally in Python.

maindlt:
When using the Python NNLS options for the weight solver:
- Calculate the projected masses only once and store them (a) in an attribute of the visible components' mge and (b) in `mass_aper.ecsv` in the output directory.
- Depending on the number of pixels in the kinematics, this results in a significant performance increase.
- The legacy Fortran programs `triaxmassbin` and `triaxmassbin_bar` are not used anymore and hence not compiled by the makefile when executing `make nogal`.

Note that when using the deprecated `LegacyWeightSolver`/GALAHAD, the old behavior is retained i.e., `triaxmassbin`/`triaxmassbin_bar` are still run for each orbit library.

**Tests part 1:**

Tested with both `test_nnls.py` and `test_bar.py` and confirmed the new projected mass calculations coincide with the legacy ones:
1. Run the test script with weight solver type `NNLS`.
2. Copy `mass_aper.ecsv` in the output directory somewhere else.
3. Run the test script with the `LegacyWeightSolver`.
4. Compare any model's `datfil/mass_aper.dat` to the saved `mass_aper.ecsv`. The first few digits of the numbers should be the same.

**Tests part 2:**

Testing the new utility method `Configuration.remove_projected_masses_file()`:
1. Run `test_nnls.py` with the Python NNLS. Verify that there is the file `mass_aper.ecsv` in the output directory.
2. In the `dev_tests/` directory, open a Python shell and enter:
```
>>> import dynamite as dyn
>>> c = dyn.config_reader.Configuration('user_test_config_ml.yaml')
>>> c.remove_projected_masses_file()
```
3. Verify that the file `mass_aper.ecsv` was deleted.


maindlt:
The last commit allows to calculate the projected masses in parallel processes.
- For higher-resolution kinematics the speedup can be significant e.g., FCC047 with 2 kinematic sets and a total of 175562 pixels, 837 spatial (Voronoi) bins, len(mge)=12:
4min 56sec sequential -> 1min 29sec parallel with ncpus=8.
- Inside parallel processing in the model iterator, the weight solver calls `mge.get_projected_masses()` with the `parallel=False` flag. This shouldn't impact performance because the projected masses are calculated once after reading the configuration file and the weight solver will get the cached projected masses without re-integration.

maindlt:
Closes #511.

prashjet:
Ran`test_nnls.py` and confirmed that:
- entries agree between `mass_aper.ecsv` (w/ `NNLS`) and `mass_aper.dat` (with `LegacyWeightSolver`),
- overall `chi2` values agree between two runs,
- the `remove_projected_masses_file` works as expected

Tested successfully with `test_notebooks`also.

Happy to merge!

PR #517, opened 2026-04-20
New intrinsic mass calculations

maindlt:
When using the Python NNLS options for the weight solver:
- Calculate the intrinsic masses in Python and store them in the respective orblib's `datfil/` directory (`mass_qgrid.ecsv`, `mass_radmass.ecsv`).
- The Python code's runtimes are 1.5-2.4 times the legacy Fortran's, but they are still negligible compared to realistic orbit library initial condition and orbit integration runtimes: in the standard `(nr, nth, nph) = (10, 6, 6)` setting for different systems, integrating the intrinsic masses takes about 6-16 seconds in Python versus 2.9-9.5 seconds in Fortran. Packages like numba could narrow this gap, but it's probably not a top priority right now.
- The legacy Fortran programs `triaxmass` and `triaxmass_bar` are not used anymore and hence not compiled by the makefile when executing `make nogal`.

Note that when using the deprecated `LegacyWeightSolver`/GALAHAD, the old behavior is retained i.e., `triaxmass`/`triaxmass_bar` are still run for each orbit library and the legacy files `mass_qgrid.dat`, `mass_radmass.dat` are created.

Tested with both `test_nnls.py` and `test_bar.py` and confirmed the new intrinsic mass calculations coincide with the legacy ones:
1. Run the test script with weight solver type `NNLS`.
2. Copy `mass_qgrid.ecsv` and `mass_radmass.ecsv` from an orblib's `datfil/` folder somewhere else.
3. Run the test script with the `LegacyWeightSolver`.
4. Compare any model's `datfil/mass_qgrid.dat ` and `mass_radmass.dat` to the previously copied files. The first few digits of the numbers should be the same.
5. The overall result of the DYNAMITE runs should make sense.

### DIFF
--- a/dev_tests/bartest.yaml
+++ b/dev_tests/bartest.yaml
@@ -196,10 +196,10 @@ orblib_settings:
     random_seed: 4242 # integer; any value <= 0 results in stochastic seed
 
 weight_solver_settings:
-    # type: "LegacyWeightSolver"
-    # nnls_solver: 1
-    type: "NNLS"
-    nnls_solver: 'scipy'
+    type: "LegacyWeightSolver"
+    nnls_solver: 1
+    # type: "NNLS"
+    # nnls_solver: 'scipy'
     maxiter_factor: 3  # nnls_solver 'scipy' only; default: 3
     CRcut: True
     regularisation: 0
--- a/dev_tests/user_test_config_ml_gas.yaml
+++ b/dev_tests/user_test_config_ml_gas.yaml
@@ -32,7 +32,6 @@ system_components:
                 LaTeX: "$a_\\mathrm{BH}$"
         type: "Plummer" # class name
         include: True
-        contributes_to_potential: True
 
     dh: # dark_halo
         parameters:
@@ -58,7 +57,6 @@ system_components:
                 LaTeX: "$\\log(M_{200}/M_\\mathrm{stars})$"
         type: NFW
         include: True
-        contributes_to_potential: True
 
     # dh: # dark_halo
     #     parameters:
@@ -74,7 +72,6 @@ system_components:
     #             LaTeX: "$\\log(M_{200}/M_\\mathrm{stars})$"
     #     type: NFW_m200_c
     #     include: True
-    #     contributes_to_potential: True
 
     stars:
         parameters:
@@ -109,7 +106,6 @@ system_components:
         mge_pot: "mge.ecsv"
         mge_lum: "mge.ecsv"
         include: True
-        contributes_to_potential: True
         kinematics: # optional, only specified for components with kinematics
             kinset1:
                 type: GaussHermite # specifies which object class to create
--- a/docs/getting_started/code_overview.rst
+++ b/docs/getting_started/code_overview.rst
@@ -267,7 +267,7 @@ Different models can be run as separate processes. The number of processes which
   multiprocessing_settings:
       ncpus: 4 # an integer or 'all_available'
 
-If ``ncpus: 'all_available'`` is selected, the program will automatically detect the total number of disposable cpus.
+If ``ncpus: 'all_available'`` is selected, the program will automatically detect the total number of disposable cpus. There is more information on the multiprocessing settings in the :ref:`multiprocessing_settings` section of the Configuration page.
 
 If you use the Slurm job submission system on a cluster, then you must add a Python `shebang line <https://en.wikipedia.org/wiki/Shebang_(Unix)>`_ and any Slurm settings to the top of ``main_script.py`` e.g.
 
--- a/docs/getting_started/code_overview.rst
+++ b/docs/getting_started/code_overview.rst
@@ -46,9 +46,10 @@ After running the script, the following directories/files will be created::
   | ├── main_script.py
   | ├── dynamite.log    # a log file
   | ├── output
-  | │   ├── models/     # output model directory
-  | │   ├── plots/      # output plot directory
-  | │   ├── all_models.ecsv        # summarises all models run so far
+  | │   ├── models/          # output model directory
+  | │   ├── plots/           # output plot directory
+  | │   ├── all_models.ecsv  # summarises all models run so far
+  | │   └── mass_aper.ecsv   # projected masses (Python NNLS only)
   |
 
 Subsequent runs of scripts from the main directory (e.g. after you have altered configuration settings) will not change this directory structure.
--- a/docs/getting_started/configuration.rst
+++ b/docs/getting_started/configuration.rst
@@ -346,6 +346,8 @@ Settings specifying the location of input and output directory names. Paths are 
         output_directory: "output/"         # directory (will be created) for output
         all_models_file: "all_models.ecsv"  # filename for the summary file of models run so far
 
+.. _multiprocessing_settings:
+
 ``multiprocessing_settings``
 ============================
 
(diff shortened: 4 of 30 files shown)


prashjet:
I confirmed that compiling with `make nogal` no longer creates `triaxmass` and `triaxmass_bar`.

Running `test_nnls` with both `NNLS/scipy` and `LegacyWeightSolver`, I confirmed that:
- `mass_radmass.ecsv` agrees (to 4 sig. figs.) with `mass_radmass.dat`,
- `mass_qgrid.ecsv` agrees (to 4 sig. figs.) with `mass_grid.dat`.

One unexpected behaviour: running `test_nnls` with `LegacyWeightSolver` produced both `.ecsv` and `.dat` files. The values in these `.ecsv` files are identical to those produced from the `NNLS/scipy` runs i.e. it looks like they're using the python code, not copying across the Fortran output. Is it necessary to have both?

maindlt:
> Is it necessary to have both?

Thanks for pointing this out! It is not necessary and is a leftover from developing. The last commit skips the `.ecsv` files when using the `LegacyWeightSolver`, replicating the legacy behavior.

One more thing to discuss: with` NNLS/scipy`, the `test_nnls.py` circles are not exactly met, even though the intrinsic masses coincide, strange... 
kinchi2_vs_model_id

prashjet:
Did a quick test to see whether the (very small) differences between the two intrinsic mass calculations could be behind the difference in chi2's in the plot above.

image

For this plot I vary the adopted error on intrinsic mass used during weight solving (x-axis) from 1% (i.e. the default value, shown by leftmost points), to 30% and 900%. The y-axis shows the  Delta chi^2 between solutions using the two intrinsic weight calculations. The three colours represent the three models in the black-circle/red-cross plot.

I anticipated that by increasing the error, any difference between the two intrinsic masses would become washed out, and delta chi_2's would approach 0. Clearly this isn't the case in general. Not sure yet what the problem is...

prashjet:
Hey @maindlt - toying around with the sum to 1 constraint didn't change things.

Some other thoughts... 

The relative changes in chi2 are of order 1%. This does feel large relative to the 10^-6% change in the intrinsic masses, but given that the condition number for this test is around 10^4, maybe a 1% change in chi2 isn't sooo unexpected. I still don't really understand why the Delta chi2 doesn't approach 0 as the error on intrinsic masses are increased (this reduces the condition number to 10^3.5). Nevertheless, having thought about it a little more, the absolute size of the difference doesn't feel so unexpected now

prashjet:
It may have been premature to say...

> Hey @maindlt - toying around with the sum to 1 constraint didn't change things.

... In the end I actually do think it may be cause of the difference. This is because:

- the new intrinsic masses sum to 1 exactly while the old ones don't,
- the uncertainty on the sum of weights constraint depends on the intrinsic masses:
   - For legacy weight solver (e.g. line 545 of `triaxnnls_noCRcut`) this uncertainty is set to the minimum of (i) a tenth of the supplied intrinsic mass error and (ii) the difference between 1 and the sum of  intrinsic masses. For our legacy run, the second option is smaller, giving an uncertainty of 5e-6 %.
   - NNLS reproduces the behaviour of legacy weight solver (line 664 of `weight_solvers.py`) but then when constructing the design matrix (line 696 of `weight_solvers.py`) additionally imposes that if this uncertainty is <=0, then we just use a 1% uncertainty instead. Since the new intrinsic masses sum exactly to 1, the resulting this means we always get a 1% uncertainty

Are the new intrinsic masses explicitly rescaled to sum exactly to 1? If so, removing this should give us the old behaviour.

maindlt:
> Are the new intrinsic masses explicitly rescaled to sum exactly to 1? If so, removing this should give us the old behaviour.

I'll check, many thanks for analyzing, @prashjet!

In the meanwhile, I have been experimenting with np.reshape(...order='F') or not (line 518 in mges.py) and strangely it doesn't matter? I'll pick that up after changing the intrinsic grid from 10, 6, 6, to differing numbers like 10, 6, 7, or so...

maindlt:
Update: you were spot-on @prashjet - the root cause is the `total_mass_error` not being zero in case of the legacy intrinsic masses, but zero in case of the new Python implementation. In `mges.py:824`, the intrinsic `quad_grid` masses are normalised by the total mass (`quad_grid /= total_mass`), but this is necessary because in `test_nnls.py` the total mass is very big (`total_mass = 16248751639.34527`) and thus *not* doing the normalisation doesn't give reasonable results. I don't know why the sum of total_mass is not exactly normalised to 1 in the Fortran implementation...

So as a fix I thought `1e-8` is a reasonably small number and added it to the total mass error in `weight_solvers.py:663` in the last commit: `self.total_mass_error = np.min([self.intrinsic_mass_error/10., np.abs(1. - self.total_mass)]) + 1e-8`. That way, zero errors are avoided and the results are the same between `LegacyWeightSolver` and `NNLS/scipy` (almost, 5e-6 would perhaps be better in this case, but what about others?). I think that's it...

One more thing that I noticed after being confused by the Fortran versus C order of indices going back and forth between Fortran an Python:

Legacy behavior:
- In both Fortran and Python, the intrinsic masses have shape `(quad_nph, quad_nth, quad_nr)`.
- Fortran flattens the array (first index changes fastest)
- Reading the intrinsic masses in `MGE.get_intrinsic_masses_from_file()` reshapes to `(quad_nph, quad_nth, quad_nr)`, BUT using the C order (last index changes fastest). This is probably ok because when reading the orbit intrinsic masses (`density_3D`), `density_3D` has its dimensions reversed (e.g., `orblib.py:786`). So weight solving should be ok in any case.

New behavior (Python intrinsic masses implementation):
- Same as above: In both Fortran and Python, the intrinsic masses have shape `(quad_nph, quad_nth, quad_nr)`.
- `mges.py:825` flattens the array with `order='F'` (first index changes fastest), mimicking legacy behavior.
- `MGE.get_intrinsic_masses()` returns the intrinsic masses shaped to `(quad_nph, quad_nth, quad_nr)`, BUT using the C order (last index changes fastest) (mges.py:200 and mges.py:517, resp.). This again, mimics legacy behavior. So all good...

prashjet:
The 3D intrinsic masses of the orbits sum to 1 exactly, so I think it is consistent that the stellar MGE intrinsic masses also sum to 1 exactly, as per the new implementation.

I think setting `self.total_mass_error` to some tiny value makes sense. I'd be inclined to make things simple with 
```
self.total_mass_error =  1e-8
```
rather than the somewhat opaque looking...
```
self.total_mass_error = np.min([self.intrinsic_mass_error/10., np.abs(1. - self.total_mass)]) + 1e-8
```
As I understand things, the weights should sum to 1 by definition i.e. this would be an equality constraint if we used an optimiser other than NNLS. So, I think there's no need to have this convoluted setting for `self.total_mass_error`... just some tiny, hard-coded number will do.

maindlt:
> there's no need to have this convoluted setting

Yes, I agree - the expression for `self.total_mass_error` is a bit convoluted and makes limited sense. Primarily that is because `self.intrinsic_mass_error` is set to a 10th of the `'lum_intr_rel_err'` configuration setting. That setting is a relative error but `self.total_mass_error` is an absolute error...

> the weights should sum to 1 by definition

Yes, but `MGE.get_intrinsic_masses()` (in `mge.py:664`) calculates the total mass that is used to normalise the intrinsic masses later (in `mge.py:1063`) via `total_mass = math.tau * sum(surf_km * qobs * sigobs_km ** 2)`. So in the (pathological? even possible?) case that the normalisation is off, the intrinsic masses won't sum up to 1.

In summary, I would propose to set the total mass error to at least a small number but also consider the (unlikely) event that the intrinsic mass normalisation is off:
`self.total_mass_error = max(abs(1. - self.total_mass), 1e-8)` (see `weight_solvers.py:647`)

The last commit implements this; what do you think, @prashjet?

prashjet:
Sounds good - will retest and merge

Issue #228, opened 2022-05-16
Distance Conversions


adriano-poci:
Currently the internal code converts the distance, given in `parameters.in` in `[Mpc]`, using a simple geometric angular conversion
```
conversion_factor = distance*1.0e6_dp*tan(pi_d/(648d3))*parsec_km
```
Compared to a proper cosmological conversion, the discrepancy is >100pc for z > 0.1

Any solution would necessarily need to either assume a cosmology, or allow for user-specified cosmological parameters. I'm guessing the latter would be preferred, but would entail modifications.

Happy for discussion on the best way forward.

maindlt:
I came across this article "Distance measures in cosmology" that may help in implementing a proper cosmological conversion.

As a note: the conversion happens in multiple places in the code (`VisibleComponent.get_M_stars_tot()`, `Plotter.orbit_plot()`, `Decomposition.create_orbital_component_files_giu()`), perhaps a system parameter would be in order...

[dist_ref.pdf](https://github.com/dynamics-of-stellar-systems/dynamite/files/10527802/dist_ref.pdf)

PR #519, opened 2026-04-22
Adopt H_0=70 throughout DYNAMITE

### LINKED ISSUE #362: Hard-coded H_0
If using the M_200 - c coupled halo option, h is set to h=0.671  km/s/Mpc (line 999 in physical_system.py). However, in iniparam_f.f90, h is set to 73 km/s/Mpc (line 80), which means it is possible for two different cosmologies to be adopted simultaneously within the code. Also, the M_200-c coupled halo relation used is only valid for z = 0 (but there are redshift dependent relations given in  Dutton & Maccio 2014), which I think should be clarified in the documentation. More broadly, the hard-coding of a H_0 value in iniparam_f.f90 is not flexible to pushing the use cases of dynamite to higher redshifts.

maindlt:
This PR adopts $H_0=70\ km/s/Mpc$ throughout DYNAMITE, a value often used in the literature.

In the documentation, an according note was added to the `Configuration -> system_attributes` section, along with mentioning that as of now, DYNAMITE systems are assumed at zero redshift.

Closes #362

This PR is ready for testing and merging.
Test suggestion: run `test_nnls.py`, `test_notebooks.sh`, and have a brief look at the changed code to verify.

### DIFF
--- a/docs/getting_started/configuration.rst
+++ b/docs/getting_started/configuration.rst
@@ -82,6 +82,8 @@ This section lists the following attributes of the system::
       distMPc: ...        # distance in MPc
       name:  ...          # name for your galaxy
 
+Note: DYNAMITE assumes :math:`H_0 = 70\;\mathrm{km/s/Mpc}` and that the system is at zero redshift :math:`z=0`. An extension to different cosmologies and system redshifts may be implemented at a later point.
+
 ``system_components``
 =====================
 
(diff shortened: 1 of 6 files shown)


prashjet:
Tested successfully with `test_nnls` and `test_notebooks` + code looks good. Happy to merge.

But let's revisit #228 again soon to implement the general solution for general cosmologies and redshifts...

PR #520, opened 2026-04-25
Check for right bar-disk system variant on startup

### LINKED ISSUE #516: Check missing bar-disk method triax_pqu2tpp_bar()
In `orblib.py: LegacyOrbitLibrary.create_fortran_input_orblib()`, bar-disk systems with parameters `q, p, u, qdisk` are expected to have a method `triax_pqu2tpp_bar()`. However, this method is not implemented in `physical_system.py`. Check and add if appropriate.

https://github.com/dynamics-of-stellar-systems/dynamite/blob/7194ac8cabfdbd340548cd9cf8ad154cfed34fb0/dynamite/orblib.py#L163

maindlt:
Check for bar-disk systems with tpp ($\theta$, $\psi$, $\phi$) angles and reject the (not implemented) q, p, u, qdisk parameters upon reading the config file.

Closes #516

Testing: this is a very small PR, running `test_bar.py` and `test_nnls.py` should suffice.

### DIFF
--- a/dev_tests/bartest.yaml
+++ b/dev_tests/bartest.yaml
@@ -136,7 +136,7 @@ system_components:
                 fixed: True
                 value: 90.0
                 LaTeX: "$\\psi$"
-        type: "BarDiskComponentAngles" # class name
+        type: "BarDiskComponent" # class name
         mge_pot: "bar_mge.ecsv"
         mge_lum: "bar_mge.ecsv"
         disk_pot: "disk_mge.ecsv"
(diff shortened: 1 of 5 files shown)


prashjet:
Is this the same as getting rid of the `BarDiskComponent` and instead just keeping the `BarDiskComponentAngles` component...?

If that's simpler, should we just do that...?

maindlt:
> If that's simpler, should we just do that...?

It would definitely be simpler, let's discuss :-)

prashjet:
I removed the un-implemented bar parameterisation. There is now only one option: `BarDiskComponent`, which used to be called `BarDiskComponentAngles`. Tested successfully with `test_nnls.py` and `test_bar.py`.

@maindlt - if you could glance over and test again, then feel free to merge

maindlt:
Many thanks Prash!

Successfully re-tested with `test_bar.py` (with both Python NNLS and LegacyWeightSolver) and `test_nnls.py` - will merge...

PR #523, opened 2026-04-26
Direct ics

maindlt:
As creating the initial conditions (ICS) is just calling one Fortran executable, this PR eliminates the intermediate step of writing and executing a shell script doing just that.
- This new implementation directly executes the ICS program, capturing its return code and in case of error displaying its message (as before).
- No extra shell script, no extra method write_executable_for_ics().

Tested with `test_nnls.py`, `test_bar.py`, and `test_notebooks.sh`.

Ready for testing and merging ;-)

### DIFF
--- a/dynamite/orblib.py
+++ b/dynamite/orblib.py
@@ -423,7 +423,7 @@ class LegacyOrbitLibrary(OrbitLibrary):
         f.close()
 
     def get_orbit_ics(self):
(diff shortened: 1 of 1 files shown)


prashjet:
Tested with `test_nnls.py` and `test_notebooks.sh` - all as expected.

Will merge :)

PR #530, opened 2026-05-28
Fixes for Python projected and intrinsic mass calculations

maindlt:
Intrinsic masses bugfix:
- Upon restarting DYNAMITE, `AllModels.update_orblib_flags()` flagged orblibs for recalculation if the intrinsic masses were calculated by the new Python code, thus causing a significant performance issue. This PR fixes it.

Projected masses improvements:
- After reading the config file, if a projected mass file exists, check whether the spatial bins in that file are compatible with those in the kinematics sets just read from the config file. If not, recalculate the projected masses.
- Cleaned up the implementation of the `VisibleComponent.mass_aper` attribute (properly defined in the component's class rather than in the config reader code).

Tested `test_notebooks.sh` and `test_nnls.py`, the latter with the standard config as well as a non-public multiple-kinematics dataset.

### DIFF
--- a/dynamite/config_reader.py
+++ b/dynamite/config_reader.py
@@ -9,6 +9,7 @@ import importlib
 import yaml
 
 from datetime import datetime, timezone
+from astropy import table
 
 import dynamite as dyn
 from dynamite import constants as const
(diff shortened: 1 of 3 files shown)


prashjet:
Tested successfully on `test_notebooks.sh` and `test_nnls.py`. Will merge!

PR #533, opened 2026-06-02
allow modelling at z>0

prashjet:
For z>0, the provided distance should be the angular diameter distance. All arcsec --> kpc conversions then follow as before. A Hubble parameter can now also be specified in the config file under `system`. This is appended to `parameters_lum/pot.in` to be passed into the orbit integrator.

Tested with `test_nnls.py` and `test_notebooks.sh`.

@maindlt - would appreciate if you could re-test and merge if successful. Note: this requires re-making the legacy fortran programs.

Adresses #228 and (re-addresses)#362.

### DIFF
--- a/dev_tests/test_bar.py
+++ b/dev_tests/test_bar.py
@@ -41,7 +41,7 @@ def run_user_test(make_comp=False):
     # c.backup_config_file(keep=3, delete_other=True)
     # c.remove_existing_plots()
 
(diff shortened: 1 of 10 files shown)


maindlt:
Great,
- tested successfully with `test_nnls.py` and `test_notebooks.sh`
- successfully added `H: 75` to `test_nnls.py`'s configuration file's `system_attributes` and verified it is used by DYNAMITE

Happy to merge :-)

